PR #1, opened 2025-04-02
Added simple main runnable

WizardOfMenlo:
Hi, I have added a simple main to test things.
As far as I can tell, there is some issue in the params setting (it seems to be trying to require a lot of PoW bits (like 60 or so)) while the arkworks implementation only uses a few.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "whir-p3"
 version = "0.1.0"
 edition = "2024"
 rust-version = "1.85"
+default-run = "main"
 
 [lints]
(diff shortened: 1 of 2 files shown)


WizardOfMenlo:
Ah maybe it's because we should be using an extension of Goldilocks instead, is the implemnetation already taking the extension?

WizardOfMenlo:
Yeah, in parameters.rs `F::bits` gives the base field bits (instead of the extension fields bits). Changing that, everything should work (but not sure how to do it properly in p3).

tcoratger (approved):

PR #4, opened 2025-04-18
Remove PrimeSubField from traits

SyxtonPrime:
Simplified the trait system by removing all references to `PrimeSubField`.

Also removed all references to `two_inv` as it's quite a bit cheaper to just use the `.halve()` function. (Both `.halve()` is cheaper than multiplication and it avoids needing to do an inversion)

I ran the test suite and nothing seems to have gone wrong. I did have to slightly rework one or two traits to depend on both `EF` and `F` instead of just `EF`.

In general I'd recommend avoiding `PrimeSubField` unless it's absolutely necessary.

The other minor change is removing `EF::PrimeSubfield::DIMENSION` (which became `F::DIMENSION`) as that's just always equal to `1` as F is the smallest field here.

### DIFF
--- a/src/domain/mod.rs
+++ b/src/domain/mod.rs
@@ -18,7 +18,7 @@ where
 {
     /// The domain defined over the base field, used for initial FFT operations.
     ///
-    /// This is useful when operating in an extension field `F`, where `F::PrimeSubfield`
+    /// This is useful when operating in an extension field `EF`, where `F`
     /// represents the base field from which the extension was built.
     pub base_domain: Option<GeneralEvaluationDomain<F>>,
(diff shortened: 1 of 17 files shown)


tcoratger:
Thanks a lot for the PR, it seems that there are just one or two small changes missing on the non-parallel cfg similar to those made on the parallel cfg :) then we are good to go.

SyxtonPrime:
Whoops my bad. Didn't realize you had parallel enabled by default.

This should fix it.

tcoratger:
Perfect, thanks lgtm

tcoratger (approved):

PR #12, opened 2025-04-22
Conversion to EF was not parallelized, saves 1s on my machine for -d 26

WizardOfMenlo:


WizardOfMenlo:
According to samply, there is still some 0.4s, 0.6s between the commit and the prove of singlethreaded work that can probably be further optimized

PR #19, opened 2025-04-24
sumcheck single: add mixed case between F and EF

tcoratger:


WizardOfMenlo:
Btw, an optimization that might be nice in the small base field case:
- Currently the sumcheck peels off one variable at a time, so to decrease by k we send k sumcheck polynomials
- The first of these sumcheck polynomials is over the base field, so it is cheaper than the others
- Instead, one can "peel off" k_1 variables at a time. This is more expensive (there are some tricks to improve it, but naively you pay something like 3^k_1 times the costs of the sumcheck), but now the computations are entirely over the base field for this polynomial, so you end up working more in the base field.

PR #28, opened 2025-04-30
committer: rm useless extension conversion

jdodinh:
I will publish benchmark results shortly.

### DIFF
--- a/src/whir/prover.rs
+++ b/src/whir/prover.rs
@@ -1,15 +1,17 @@
 use p3_commit::{ExtensionMmcs, Mmcs};
 use p3_dft::TwoAdicSubgroupDft;
 use p3_field::{ExtensionField, Field, PrimeField64, TwoAdicField};
-use p3_matrix::{
-    dense::{DenseMatrix, RowMajorMatrix},
-    extension::FlatMatrixView,
-};
-use p3_merkle_tree::{MerkleTree, MerkleTreeMmcs};
+use p3_matrix::dense::RowMajorMatrix;
+use p3_merkle_tree::MerkleTreeMmcs;
 use p3_symmetric::{CryptographicHasher, PseudoCompressionFunction};
 use serde::{Deserialize, Serialize};
 
-use super::{WhirProof, committer::Witness, parameters::WhirConfig, statement::Statement};
+use super::{
+    WhirProof,
+    committer::{CommitmentMerkleTree, RoundMerkleTree, Witness},
+    parameters::WhirConfig,
+    statement::Statement,
+};
 use crate::{
     domain::Domain,
     fiat_shamir::{errors::ProofResult, pow::traits::PowStrategy, prover::ProverState},
@@ -44,11 +46,10 @@ where
     pub(crate) coefficients: CoefficientStorage<F, EF>,
 
     /// Prover data for the commitment is over the base field
-    pub(crate) commitment_merkle_prover_data: MerkleTree<F, u8, DenseMatrix<F>, DIGEST_ELEMS>,
+    pub(crate) commitment_merkle_prover_data: CommitmentMerkleTree<F, DIGEST_ELEMS>,
     /// Prover data for the remaining rounds is over the extension field
     /// None in the first round, or if the number of rounds is zero
-    pub(crate) merkle_prover_data:
-        Option<MerkleTree<F, u8, FlatMatrixView<F, EF, DenseMatrix<EF>>, DIGEST_ELEMS>>,
+    pub(crate) merkle_prover_data: Option<RoundMerkleTree<F, EF, DIGEST_ELEMS>>,
     /// Merkle proofs
     /// - The first element is the opened leaf values
     /// - The second element is the Merkle proof (siblings)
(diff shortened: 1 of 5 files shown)


jdodinh:
Running the `whir_p3::whir::make_whir_things` test using Samply yields the following results on my machine (M3 Mac):
**Before the change:**
Total time to run `whir_p3::whir::make_whir_things`: 10.0 seconds (7139 samples)
52% spent in `whir_p3::whir::prover::Prover::prove` (3687 samples)
24% spent in `whir_p3::whir::verifier::Verifier::verify` (1693 samples)
11% spent in `whir_p3::whir::committer::writer::CommitmentWriter::commit` (790 samples)
Screenshot 2025-04-30 at 16 10 24

**After the change:**
Total time to run `whir_p3::whir::make_whir_things`: 7.9 seconds (7381 samples)
51% spent in `whir_p3::whir::prover::Prover::prove` (3753 samples)
22% spent in `whir_p3::whir::verifier::Verifier::verify` (1660 samples)
13% spent in `whir_p3::whir::committer::writer::CommitmentWriter::commit` (978 samples)
Screenshot 2025-04-30 at 16 07 58

jdodinh:
I'm attaching here Sample runs that you can upload to [https://profiler.firefox.com/](the profiler).

[Uploading after_1.json…]()

src/whir/prover.rs
```diff
@@ -282,25 +293,57 @@ where
...
-            merkle_proof.push(proof);
-        }
+        let stir_evaluations = match &round_state.merkle_prover_data {
+            None => {
+                let mut answers = Vec::new();
+                let mut merkle_proof = Vec::new();
+                let commitment_merkle =
+                    MerkleTreeMmcs::new(self.0.merkle_hash.clone(), self.0.merkle_compress.clone());
+                for challenge in &stir_challenges_indexes {
+                    let (commitment_leaf, commitment_root): (Vec<Vec<F>>, Vec<[u8; DIGEST_ELEMS]>) =
```
tcoratger, line 303:
```suggestion
                    let (commitment_leaf, commitment_root) =
```

src/whir/prover.rs
```diff
@@ -282,25 +293,57 @@ where
...
+                        commitment_merkle
+                            .open_batch(*challenge, &round_state.commitment_merkle_prover_data);
+                    answers.push(commitment_leaf[0].clone());
+                    merkle_proof.push(commitment_root);
+                }
+                // Evaluate answers in the folding randomness.
+                let mut stir_evaluations = ood_answers;
+                let transformed: Vec<Vec<EF>> = answers
+                    .iter()
+                    .map(|inner| inner.into_iter().map(|&fel| EF::from(fel)).collect())
```
tcoratger, line 313: I think that we can keep that for a follow up but usually we should not due such a conversion because all the algebra is implemented for extension field with base field, so that it is usually better to keep things in the base field as much as we can.

src/whir/prover.rs
```diff
@@ -282,25 +293,57 @@ where
...
-            self.0.folding_factor,
-            &mut stir_evaluations,
-        );
-        round_state.merkle_proofs.push((answers, merkle_proof));
+                round_state.commitment_merkle_proof = Some((answers, merkle_proof));
+                stir_evaluations
+            }
+            Some(data) => {
+                let mut answers = Vec::new();
+                let mut merkle_proof = Vec::new();
```
tcoratger, line 328: Sam here can't we init with capacity?

src/whir/prover.rs
```diff
@@ -392,15 +435,34 @@ where
...
-            merkle_proof.push(proof);
-        }
+        match &round_state.merkle_prover_data {
+            None => {
+                let mut answers = Vec::new();
+                let mut merkle_proof = Vec::new();
+                let commitment_merkle =
+                    MerkleTreeMmcs::new(self.0.merkle_hash.clone(), self.0.merkle_compress.clone());
+                for challenge in final_challenge_indexes {
+                    let (commitment_leaf, commitment_root): (Vec<Vec<F>>, Vec<[u8; DIGEST_ELEMS]>) =
```
tcoratger, line 445:
```suggestion
                    let (commitment_leaf, commitment_root) =
```

src/whir/prover.rs
```diff
@@ -392,15 +435,34 @@ where
...
-        let mut merkle_proof = Vec::new();
-        for challenge in final_challenge_indexes {
-            let (leaf, proof) = mmcs.open_batch(challenge, &round_state.prev_merkle_prover_data);
-            answers.push(leaf[0].clone());
-            merkle_proof.push(proof);
-        }
+        match &round_state.merkle_prover_data {
+            None => {
+                let mut answers = Vec::new();
+                let mut merkle_proof = Vec::new();
```
tcoratger, line 441: Can't we init with capacity?

src/whir/verifier.rs
```diff
@@ -145,15 +149,42 @@ where
...
+                            &prev_root,
+                            &dimensions,
+                            stir_challenges_index,
+                            &[indexed_answers],
+                            &merkle_proof[i],
+                        )
+                        .map_err(|_| ProofError::InvalidProof)?;
+                    stir_challenges_answers = answers
+                        .into_iter()
+                        .map(|inner| inner.into_iter().map(|&f_el| EF::from(f_el)).collect())
```
tcoratger, line 174: Same here we can keep for later as the scope of this PR is already sufficient but this conversion should be avoided for performance reasons.

tcoratger:
@jdodinh Thanks a lot for the PR, this is exactly what we want.
- I just did a first pass with a few minor comments.
- You can run the clippy locally to fix the small issues.

jdodinh:
Just addressed most of the comments. Regarding the conversions, I agree that we should avoid them as much as possible. The improvement is noticeable because we are only converting the necessary leaves, rather than the entire polynomial evaluation, but I think there is a way we can avoid it altogether. 

Also, a refactoring will be necessary, since currently there is a lot of repeated code

Would you like this to remain as part of this PR, or should we create a separate issue out of it?

PR #38, opened 2025-05-06
sumcheck: implement univariate skip

tcoratger:
There are a lot of optimizations to do on this implementation but for now I just wanted to get the core logic correct, @WizardOfMenlo @SyxtonPrime can you take a look to check if it corresponds to what you have in mind?

To guide you through the PR here is a description of the logic made.

## Motivation:

This PR introduces the **Univariate Skip** optimization for our sumcheck protocol, aiming to improve prover performance. Based on the protocol described in https://eprint.iacr.org/2024/108.pdf and https://hackmd.io/@tcoratger/rJOj6L1egx, this optimization strategically shifts computational effort. Standard sumcheck incurs significant cost due to extension field operations in rounds $1$ to $n-1$. The Univariate Skip technique performs the work equivalent to the first `k` rounds primarily within the base field $F$, thereby reducing the reliance on more expensive extension field $G$ arithmetic.

## Implementation details:

The core mechanism involves modifying the initial phase of the sumcheck protocol when `k_skip = Some(k)` (where $k \ge 2$) is provided:

1.  **Prover's skip round:**
    * Instead of computing the standard quadratic round polynomial, the prover computes evaluations of a higher-degree *univariate* polynomial $v_0(X) = \sum_{\mathbf{x} \in H^{n-k}} \omega_{n-k}(\boldsymbol{\alpha}, \mathbf{x}) f(X, \mathbf{x})$.
    * This is achieved in `compute_skipping_sumcheck_polynomial` by:
        * Performing Low-Degree Extensions (LDE) of the base polynomial evaluations (`p`) and the constraint weights (`w`) onto a multiplicative coset of size $2^{k+1}$.
        * Computing the sum of the pointwise products $p(X, \mathbf{x}) \cdot w(X, \mathbf{x})$ over the remaining variables $\mathbf{x}$ for each point $X$ on the coset.
    * The resulting $2^{k+1}$ evaluations of $v_0(X)$ are added to the Fiat-Shamir transcript.

2.  **Challenge handling & state update:**
    * The prover obtains $k$ challenges $r_0, \dots, r_{k-1}$ from the transcript.
    * The prover simulates the verifier's check by evaluating the committed polynomial $v_0(X)$ at the first challenge $r_0$. This is done using the `evaluate_univariate_poly_at_challenge` function (we could probably use something else already existing but I've isolated the logic here to make things easy to check and modify without any other breaking change), which performs an iDFT on the committed evaluations and uses Horner's method. The result $v_0(r_0)$ becomes the claimed sum for the next phase.
    * The prover updates its internal state by folding the *original* base field polynomial evaluations and weights $k$ times using all challenges $r_0, \dots, r_{k-1}$ via the `fold_k_times` helper. The state now involves extension field elements and $k$ fewer variables.

3.  **Subsequent rounds:**
    * The protocol proceeds with the standard sumcheck rounds (`compute_sumcheck_polynomial` generating quadratic polynomials, followed by `compress`) for the remaining `folding_factor - k` variables, operating on the already-folded state.

---

### DIFF
--- a/src/sumcheck/sumcheck_single_skip.rs
+++ b/src/sumcheck/sumcheck_single_skip.rs
@@ -1,0 +1,636 @@
+use p3_dft::TwoAdicSubgroupDft;
+use p3_field::{ExtensionField, Field, TwoAdicField};
+use p3_matrix::{Matrix, dense::RowMajorMatrix};
+#[cfg(feature = "parallel")]
+use rayon::iter::{IndexedParallelIterator, ParallelIterator};
+
+use super::{sumcheck_polynomial::SumcheckPolynomial, sumcheck_single::SumcheckSingle};
+use crate::poly::evals::EvaluationStorage;
+
+impl<F, EF> SumcheckSingle<F, EF>
+where
+    F: Field + TwoAdicField,
+    EF: ExtensionField<F>,
+{
+    /// Computes the sumcheck polynomial using the **univariate skip** optimization.
+    ///
+    /// This method folds the first `k` variables in one step, skipping `k` rounds of sumcheck.
+    /// The result is a univariate polynomial over a multiplicative coset of size `2^{k+1}`.
+    ///
+    /// # Input
+    /// - `dft`: A DFT implementation over the base field `F`.
+    /// - `k`: The number of initial variables to skip (fold) in one shot.
+    ///
+    /// # Output
+    /// - A degree-`2^k - 1` univariate polynomial over the extension field `EF`,
+    ///   returned as a `SumcheckPolynomial<EF>`.
+    ///
+    /// # Constraints
+    /// - This function assumes that the original polynomial and weights are defined over the **base field**.
+    /// - It panics if the polynomial is already defined over the extension field.
+    pub fn compute_skipping_sumcheck_polynomial<DFT>(
+        &self,
+        dft: &DFT,
+        k: usize,
+    ) -> SumcheckPolynomial<EF>
+    where
+        DFT: TwoAdicSubgroupDft<F>,
+    {
+        // Ensure we have at least `k` variables to skip.
+        // Otherwise, univariate skip is not applicable.
+        assert!(
+            self.num_variables() >= k,
+            "Need at least k variables to apply univariate skip on k variables"
+        );
+
+        // Evaluate based on the storage format of the polynomial:
+        // Only base field evaluations are supported for skipping.
+        let out_vec = match &self.evaluation_of_p {
+            EvaluationStorage::Base(evals_f) => {
+                // Interpret polynomial evaluations as a 2D matrix
+                let f_mat = RowMajorMatrix::new(evals_f.evals().to_vec(), 1 << k).transpose();
+                let weights_mat =
+                    RowMajorMatrix::new(self.weights.evals().to_vec(), 1 << k).transpose();
+
+                // Choose a 2-adic generator of order 2^{k+1}, which defines a coset D of size 2^k.
+                // This lets us extend from domain {0,1}^k to a multiplicative domain of size 2^{k+1}.
+                let two_adic_gen = F::two_adic_generator(k + 1);
+
+                // Apply low-degree extension (LDE) to both:
+                // - Coefficients matrix (over F)
+                // - Weights matrix (over EF)
+                //
+                // This evaluates both over the multiplicative coset domain D.
+                let f_on_coset = dft
+                    .coset_lde_batch(f_mat, 1, two_adic_gen)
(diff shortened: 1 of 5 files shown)


src/sumcheck/sumcheck_single_skip.rs
```diff
@@ -0,0 +1,636 @@
...
+                // - Coefficients matrix (over F)
+                // - Weights matrix (over EF)
+                //
+                // This evaluates both over the multiplicative coset domain D.
+                let f_on_coset = dft
+                    .coset_lde_batch(f_mat, 1, two_adic_gen)
+                    .to_row_major_matrix();
+                let weights_on_coset = dft
+                    .coset_lde_algebra_batch(weights_mat, 1, two_adic_gen)
+                    .to_row_major_matrix();
```
SyxtonPrime, line 69: You probably? want to use `lde_batch?` (or equivalently set the shift to `1`?)

(I think in the code I originally sent you, I had added bits as `0` as can use the `f_mat` and `weights_mat` inputs and just merge the matrices together later as a minor optimization)

But in the short term, while this is extrapolating onto a large domain, I think we want to get the values on the subgroup of order `2^{k+1}` in the standard order.

src/sumcheck/sumcheck_single_skip.rs
```diff
@@ -0,0 +1,636 @@
...
+        );
+
+        // Evaluate based on the storage format of the polynomial:
+        // Only base field evaluations are supported for skipping.
+        let out_vec = match &self.evaluation_of_p {
+            EvaluationStorage::Base(evals_f) => {
+                // Interpret polynomial evaluations as a 2D matrix
+                let f_mat = RowMajorMatrix::new(evals_f.evals().to_vec(), 1 << k).transpose();
+                let weights_mat =
+                    RowMajorMatrix::new(self.weights.evals().to_vec(), 1 << k).transpose();
```
SyxtonPrime, line 53: Are the weights definitely going to be in the extension field at this point?

(Would be much nicer if they where in the base field but that may well not be possible)

src/sumcheck/sumcheck_single_skip.rs
```diff
@@ -0,0 +1,636 @@
...
+        "Coefficient count mismatch after iDFT"
+    );
+
+    // Evaluate the polynomial P(X) = sum_{i=0}^{N-1} coeffs_v[i] * X^i at X = challenge_r0
+    // Use Horner's method for evaluation.
+    let mut value = EF::ZERO;
+    // Iterate from the highest degree coefficient down to the constant term
+    for coeff in coeffs_v.iter().rev() {
+        value = value * challenge_r0 + *coeff;
+    }
```
SyxtonPrime, line 165: Hmm, just an optimization comment.

This type of interpolation can be done more quickly by avoiding the FFT and using a lagrange interpolation approach.

Check out Plonky3's `interpolate_coset` function. (Or more generally all functions inside `interpolation/src/lib`

SyxtonPrime (commented):
Hmm, this all looks good to me but I'm a little confused why the prover needs to do both the univariate skip but then still `fold_k_times`?

I guess I assumed the verifier could just verify the `univariate_skip` step.

src/sumcheck/sumcheck_single_skip.rs
tcoratger: Mhhh, okay, it seemed weird to me to have this shift too, I thought I must have something wrong, so it should be this right?

```rust
// Apply low-degree extension (LDE) to both:
// - Coefficients matrix (over F)
// - Weights matrix (over EF)
//
// This evaluates both over the multiplicative coset domain D.
let f_on_coset = dft.lde_batch(f_mat, 1).to_row_major_matrix();
let weights_on_coset = dft.lde_algebra_batch(weights_mat, 1).to_row_major_matrix();
```

So that, for example, if in the batch we initially had two polynomials:
```text
f_0(x) = 1 + 2x + 3x^2 + 4x^3
f_1(x) = 5 + 6x + 7x^2 + 8x^3
```

that are each evaluated on the 4-point subgroup H = { ω^0, ω^1, ω^2, ω^3 } where ω is a primitive 4th root of unity, then we should end up with an extension of these evaluations to the 8-point domain (K = { ω^0, ω^1, ..., ω^7 } where ω is a primitive 8th root), correct?

src/sumcheck/sumcheck_single_skip.rs
tcoratger: Ah nice will take a note of this to mark this as an optimization for a followup PR to keep the flow of PRs and avoid keeping this opened for ages :)

PR #39, opened 2025-05-06
reduce proof size by sending coeffs earlier

TomWambsgans:
As soon as we reach a number of variables <= K in the WHIR steps, we stop folding and send directly the coefficients of the polynomial. In the current implem, K = folding_factor - 1 . But in practice folding_factor is often = 4, and we could send the coeffs earlier, ie use a greater value for K.

In my PR K is called "MAX_NUM_VARIABLES_TO_SEND_COEFFS" (TODO find a better name)
Ideally K should not be harcoded, because the optimal value depends on several factors.

```
cargo run --release --package whir-p3 --bin main -- --security-level=128 --num-variables=22 --sec=JohnsonBound
```

Before:
Before: 278.6 KiB
After: 265.8 KiB

### DIFF
--- a/src/parameters/mod.rs
+++ b/src/parameters/mod.rs
@@ -11,6 +11,9 @@ use crate::whir::{
 
 pub mod errors;
 
+/// Each WHIR steps folds the polymomial, which reduces the number of variables.
+/// As soon as the number of variables is less than or equal to `MAX_NUM_VARIABLES_TO_SEND_COEFFS`,
+/// the prover sends directly the coefficients of the polynomial.
 const MAX_NUM_VARIABLES_TO_SEND_COEFFS: usize = 6;
 
 /// Computes the default maximum proof-of-work (PoW) bits.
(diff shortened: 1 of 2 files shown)


src/parameters/mod.rs
```diff
@@ -259,37 +261,37 @@ impl FoldingFactor {
...
-                let final_sumcheck_rounds = num_variables % factor;
-
-                // Compute the number of WHIR rounds by subtracting the final sumcheck rounds
-                // and dividing by the folding factor. The -1 accounts for the fact that the last
-                // round does not require another folding.
-                (
-                    (num_variables - final_sumcheck_rounds) / factor - 1,
-                    final_sumcheck_rounds,
-                )
+            FoldingFactor::Constant(factor) => {
```
tcoratger, line 264:
```suggestion
            Self::Constant(factor) => {
```

src/parameters/mod.rs
```diff
@@ -259,37 +261,37 @@ impl FoldingFactor {
...
+                // number of variables is less of equal than `MAX_NUM_VARIABLES_TO_SEND_COEFFS`, we stop folding and the
+                // prover sends directly the coefficients of the polynomial.
+                let num_rounds =
+                    (num_variables - MAX_NUM_VARIABLES_TO_SEND_COEFFS).div_ceil(*factor);
+                let final_sumcheck_rounds = num_variables - num_rounds * factor;
+                // The -1 accounts for the fact that the last round does not require another folding.
+                (num_rounds - 1, final_sumcheck_rounds)
             }
-            Self::ConstantFromSecondRound(first_round_factor, factor) => {
+            FoldingFactor::ConstantFromSecondRound(first_round_factor, factor) => {
```
tcoratger, line 279:
```suggestion
            Self::ConstantFromSecondRound(first_round_factor, factor) => {
```

src/parameters/mod.rs
```diff
@@ -11,6 +11,8 @@ use crate::whir::{
 
 pub mod errors;
 
+const MAX_NUM_VARIABLES_TO_SEND_COEFFS: usize = 6;
```
tcoratger, line 14: can we have a small header doc for this to indicate what this corresponds too? :)

tcoratger:
Thanks a lot for the PR, just small nits and then we are good to merge :)

tcoratger (approved):

PR #41, opened 2025-05-07
add `MixedFieldSlice` to keep `stir_evaluations_prover` in `F`

tcoratger:


src/whir/stir_evaluations.rs
```diff
@@ -77,7 +78,14 @@ where
     /// - `answers`: Oracle values — either raw evaluations (naive) or preprocessed coefficients
     ///   (prover helps).
     /// - `stir_evaluations`: Output vector where the results will be appended.
-    pub(crate) fn evaluate(&self, answers: &[Vec<F>], stir_evaluations: &mut Vec<F>) {
+    pub(crate) fn evaluate<SF>(
```
SyxtonPrime, line 81: Is `MixedFieldSlice` really necessary here? It seems like adding the `<SF>` parameter to evaluate and having `answers: &[SF]` should work.

PR #75, opened 2025-05-19
Wavelet Transform Refactor

SyxtonPrime:
This doesn't seem to give any speed ups currently but makes working with the wavelet transformation much more similar to working with FFT's.

Currently, I just defined a struct but, if we want to investigate different wavelet transforms which could be faster in different situations, we could also restructure this as a trait identical to `TwoAdicSubgroupDFT`.

I also did a minor rework on converting between Coefficient and Evaluation vectors. As this conversion is linear, it's worth attempting to do in a subfield whenever possible as this leads to better SIMD instructions. It's also good as it highlights what can be a somewhat expensive operation. (Having .into() being a large cost can be easy to miss).


tcoratger (approved):
Amazing, looks really nice

PR #76, opened 2025-05-19
Better timing info

SyxtonPrime:
I ran through and updated the end-to-end example in src/bin/main with tracing info. This should hopefully make it easier to spot optimization places. (In the very short term it probably slows things down slightly as timing takes time but long term should be worth it and we can remove the timing data as we optimise things).

In particular I've gone through the prove/commit functions and broken down the different parts to a bunch of base functions which each take up between 1-5% of the run time. Nicely, there is nothing there anymore which seems to be a massive time sink but there are a few small functions (`accumulate, compress, evaluate, add_new_equality)` which it would be worth taking a closer look at. (I set things which I checked but ended up being really small to debug so you can toggle them on/off using RUST_FLAGS).

While doing this I found a simple rewrite of something in commit which gave about a 15% speed up (and as it's a 2 line rewrite I included it here). Basically, `.resize()` is surprisingly slow when dealing with field elements. It's much much faster (E.g. like 3x faster) to create a new vector with `zero_vec` and copy data across instead of cloning a vector and extending it. I had a quick look through the rest of the code base and removed other instances of resize which popped up in timing data as well.

(This will conflict slightly with the other PR but it should be easy to fix)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,14 @@ p3-maybe-rayon = { git = "https://github.com/Plonky3/Plonky3.git", rev = "86666b
 p3-interpolation = { git = "https://github.com/Plonky3/Plonky3.git", rev = "86666b4" }
 itertools = "0.14"
 thiserror = "2.0"
+tracing = { version = "0.1.37", default-features = false, features = [
+    "attributes",
+] }
+tracing-forest = "0.1.6"
+tracing-subscriber = { version = "0.3.17", default-features = false, features = [
+    "alloc",
+    "env-filter",
+] }
 rand = "0.9"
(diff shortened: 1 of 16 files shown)


tcoratger:
Amazing, I never noticed that the resize was particularly slow, this is a nice fix, thanks a lot.

Can you just fix the small conflicts and then we are good to merge?

tcoratger:
Also I think that once we are able to use the univariate skip all around the code here https://github.com/tcoratger/whir-p3/pull/65, the `compress`, and maybe `evaluate` functions will be less used because they will intervene on the unskipped rounds and so at the end of the protocol for small dimensions only.

tcoratger (approved):

PR #86, opened 2025-05-21
Unrolling `eval_eq` a little.

SyxtonPrime:
Unrolling some of the inner layers of eval_eq gives almost a `50%` speed up!

There are some other ideas definitely worth trying here but thought I'd make a PR for the easiest one first.


tcoratger (approved):
Amazing, lgtm

PR #92, opened 2025-05-22
Use a better packing and parallelism strategy for `eval_extension`.

SyxtonPrime:
We make 2 improvements to `eval_extension`.

We change the parallelism strategy to make use of `par_chunks_exact` as opposed to the divide and conquer strategy. This should reduce the overhead a little.

We also update the non parallelism method to use PackedFieldExtensions as much as possible.

Together this gives a `~40%` speed up when dealing with Goldilocks and the degree `2` extension and a `~60%` speed up for BabyBear and the degree `5` extension.

### DIFF
--- a/src/poly/coeffs.rs
+++ b/src/poly/coeffs.rs
@@ -165,7 +165,7 @@ where
     #[instrument(skip_all, fields(size = point.num_variables()))]
     pub fn evaluate_at_extension<EF: ExtensionField<F>>(&self, point: &MultilinearPoint<EF>) -> EF {
         assert_eq!(self.num_variables, point.num_variables());
-        eval_extension_para(&self.coeffs, &point.0)
+        eval_extension_par(&self.coeffs, &point.0)
     }
 }
 
@@ -310,7 +310,7 @@ impl<F> CoefficientList<F> {
 /// ```
 /// where the product is now only over the first `L` variables.
 #[inline]
-fn eval_extension_para<F, E>(coeff: &[F], eval: &[E]) -> E
+fn eval_extension_par<F, E>(coeff: &[F], eval: &[E]) -> E
 where
     F: Field,
     E: ExtensionField<F>,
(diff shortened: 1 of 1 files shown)


src/poly/coeffs.rs
```diff
@@ -164,7 +165,7 @@ where
     #[instrument(skip_all)]
     pub fn evaluate_at_extension<EF: ExtensionField<F>>(&self, point: &MultilinearPoint<EF>) -> EF {
         assert_eq!(self.num_variables, point.num_variables());
-        eval_extension(&self.coeffs, &point.0, EF::ONE)
+        eval_extension_para(&self.coeffs, &point.0)
```
tcoratger, line 168: Maybe this renaming is better to keep aligned with the rayon namings?

```suggestion
        eval_extension_par(&self.coeffs, &point.0)
```

src/poly/coeffs.rs
```diff
@@ -290,43 +291,164 @@ impl<F> CoefficientList<F> {
...
+/// Recursively evaluates a multilinear polynomial at an extension field point.
+///
+/// Given `coeffs` in lexicographic order, this computes:
+/// ```ignore
+/// eval_poly(X_0, ..., X_n) = sum(coeffs[i] * product(X_j for j in S(i)))
+/// ```
+/// where `S(i)` is the set of variables active in term `i` (based on its binary representation).
+///
+/// - Small cases are passed to `eval_extension_unpacked`.
+/// - For larger cases, we pack extension field elements into PackedFieldExtension elements and
```
tcoratger, line 357:
```suggestion
/// - For larger cases, we pack extension field elements into `PackedFieldExtension` elements and
```

SyxtonPrime:
This code ended up a little uglier than I had initially hoped. There might be a way to clean it up but it seems like trying to make use of both parallelization and packings leads to nasty code.

tcoratger (approved):
This is totally fine, thanks, it's super well documented and the performance improvement is amazing.

PR #99, opened 2025-05-23
Further Work on Eval Eq

SyxtonPrime:
Updated Eval Eq to make use of Packings.

Unfortunately, the data on this seems to be a little mixed. For `BabyBear`, with extension field of degree 5, this gives around a `30-40%` speed up.

But for Goldilocks, with the extension field of degree `2`, its actually a little slower `10-15%` or so. This is probably due to the extra data fiddling involved and also the fact that `PackedGoldilocksAVX2` just isn't much of an improvement over the base Goldilocks. (Addition is a little faster but multiplication is mostly unchanged). Overall a little frustrating.

I've got a couple of idea's for how to speed this up a little more but it involves changes to Plonky3 so I'll have to test/merge them first over there.

Let me know if you'd rather not merge this for now.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -23,10 +23,10 @@ use whir_p3::{
     },
 };
 
-type _F = Goldilocks;
-type _EF = BinomialExtensionField<_F, 2>;
-type F = BabyBear;
-type EF = BinomialExtensionField<F, 5>;
+type F = Goldilocks;
+type EF = BinomialExtensionField<F, 2>;
+type _F = BabyBear;
+type _EF = BinomialExtensionField<_F, 5>;
 type ByteHash = Blake3;
 type FieldHash = SerializingHasher<ByteHash>;
 type MyCompress = CompressionFunctionFromHasher<ByteHash, 2, 32>;
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -1,4 +1,5 @@
 use clap::Parser;
+use p3_baby_bear::BabyBear;
 use p3_blake3::Blake3;
 use p3_dft::Radix2DitParallel;
 use p3_field::{PrimeCharacteristicRing, extension::BinomialExtensionField};
@@ -22,8 +23,10 @@ use whir_p3::{
     },
 };
 
-type F = Goldilocks;
-type EF = BinomialExtensionField<F, 2>;
+type _F = Goldilocks;
+type _EF = BinomialExtensionField<_F, 2>;
+type F = BabyBear;
+type EF = BinomialExtensionField<F, 5>;
 type ByteHash = Blake3;
 type FieldHash = SerializingHasher<ByteHash>;
 type MyCompress = CompressionFunctionFromHasher<ByteHash, 2, 32>;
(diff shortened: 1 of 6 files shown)


src/bin/main.rs
```diff
@@ -24,6 +25,8 @@ use whir_p3::{
 
 type F = Goldilocks;
 type EF = BinomialExtensionField<F, 2>;
+type _F = BabyBear;
+type _EF = BinomialExtensionField<_F, 5>;
```
tcoratger, line 29: Do we need these or maybe you did a test with them? If needed I can in a followup rename 
```rust
type F = Goldilocks;
type EF = BinomialExtensionField<F, 2>;
```

to 

```rust
type G = Goldilocks;
type EG = BinomialExtensionField<G, 2>;
```

seems like more logical if we need to mix with another field

src/utils.rs
```diff
@@ -10,7 +15,166 @@ use p3_field::Field;
 /// ```
 ///
 /// where `z_i` are the constraint points.
-pub(crate) fn eval_eq<F: Field>(eval: &[F], out: &mut [F], scalar: F) {
+#[inline]
+pub(crate) fn eval_eq<F: Field, EF: ExtensionField<F>>(eval: &[EF], out: &mut [EF], scalar: EF) {
+    // Number of threads to spawn.
+    // Long term this should be a modifiable parameter.
+    // I've chosen 32 here as my machine has 20 logical cores. PLausibly we might want to hard code this
```
tcoratger, line 22:
```suggestion
    // I've chosen 32 here as my machine has 20 logical cores. Plausibly we might want to hard code this
```

src/utils.rs
```diff
@@ -21,62 +185,129 @@ pub(crate) fn eval_eq<F: Field>(eval: &[F], out: &mut [F], scalar: F) {
...
+/// Unlike [`eval_eq_basic`], this function makes heavy use of packed values to speed up computations.
+/// In particular `scalar` should be passed in as a packed value coming from [`packed_eq_poly`].
+///
+/// Essentially using packings this functions computes
+///
+/// ```text
+/// eq(X) = scalar[j] * ∏ (1 - X_i + 2X_i z_i)
+/// ```
+///
+/// for a collection of `i` at the same time. Here `scalar[j]` should be though of as evaluations of an equality
```
tcoratger, line 249:
```suggestion
/// for a collection of `i` at the same time. Here `scalar[j]` should be thought of as evaluations of an equality
```

tcoratger:
> Updated Eval Eq to make use of Packings.
> 
> Unfortunately, the data on this seems to be a little mixed. For `BabyBear`, with extension field of degree 5, this gives around a `30-40%` speed up.
> 
> But for Goldilocks, with the extension field of degree `2`, its actually a little slower `10-15%` or so. This is probably due to the extra data fiddling involved and also the fact that `PackedGoldilocksAVX2` just isn't much of an improvement over the base Goldilocks. (Addition is a little faster but multiplication is mostly unchanged). Overall a little frustrating.
> 
> I've got a couple of idea's for how to speed this up a little more but it involves changes to Plonky3 so I'll have to test/merge them first over there.
> 
> Let me know if you'd rather not merge this for now.

@SyxtonPrime I just noted some very minor comments. Let us merge this, I like making incremental changes and it is of course not a problem if we have to go back to it after some changes made on Plonky3, on the contrary this is very good in my opinion

src/bin/main.rs
SyxtonPrime: It was mostly for testing. (It's handy to be able to test with both Golldilocks and BabyBear fields). Happy to get rid of this or rename them.

PR #145, opened 2025-06-01
Update tests to sumcheck_polynomial.rs

0xKarl98:


src/sumcheck/sumcheck_polynomial.rs
```diff
@@ -267,4 +329,18 @@ mod tests {
...
+        let evaluations: Vec<_> = (0..3_usize.pow(n_vars.try_into().unwrap()))
+            .map(|i| F::from_u64(i as u64 % 1000))
+            .collect();
+        let poly = SumcheckPolynomial::new(evaluations, n_vars);
+
+        let start = std::time::Instant::now();
+        let _sum = poly.sum_over_boolean_hypercube();
+        let duration = start.elapsed();
+        assert!(duration.as_millis() < 100); // Should be fast  
+    }
```
tcoratger, line 345: I see the objective of this test but I would remove it from the test suite here because it is dependent on the device where we run the tests, sometimes some unexpected thing can lead to the test to be a little bit slow and make the CI red for no reason, what do you think? Would remove this test for now :)

PR #147, opened 2025-06-02
Optimization

0xKarl98:


src/whir/committer/reader.rs
```diff
@@ -79,17 +79,17 @@ where
...
+        // Optimization : We can do this to prevent zero vec allocation when OOD sample is equal to 0
+        let (ood_points, ood_answers) = if ood_samples > 0 {
+            let mut points = EF::zero_vec(ood_samples);
+            let mut answers = EF::zero_vec(ood_samples);
+            verifier_state.fill_challenge_scalars(&mut points)?;
+            verifier_state.fill_next_scalars(&mut answers)?;
+            (points, answers)
+        } else {
+            (EF::zero_vec(ood_samples), EF::zero_vec(ood_samples))
+        };
```
tcoratger, line 92: Unfortunately I would love this to be true but even is the case where `ood_samples` is 0, we must have `ood_points` and `ood_answers` be vectors full of zeros, so in both cases it comes to the same thing I think. And I think that for the same thing, the new code has a little bit more lines so I would be in favor not to do this for now.

PR #159, opened 2025-06-06
fiat shamir: introduce Plonky3 challenger

tcoratger:


src/fiat_shamir/sho.rs
```diff
@@ -92,7 +94,10 @@ where
     pub fn squeeze(&mut self, output: &mut [U]) -> Result<(), DomainSeparatorMismatch> {
         match self.stack.pop_front() {
             Some(Op::Squeeze(length)) if output.len() <= length => {
-                self.ds.squeeze_unchecked(output);
+                // self.ds.squeeze_unchecked(output);
```
tcoratger, line 97:
```suggestion
```

PR #191, opened 2025-06-25
utils: use `CanSampleBits` for `get_challenge_stir_queries`

tcoratger:
Following the proposition of @kilic in #190, `CanSampleBits` trait is used to sample queries.

@WizardOfMenlo I'm not sure I understand the effect of the number of bits sampled here (`challenger.sample_bits(8)`). In the original implementation here https://github.com/WizardOfMenlo/whir/blob/9d7a97b169542f8d0787db482ee58f7b9b660848/src/whir/utils.rs#L72, we represent the queries as `u8` which is equivalent to sample over 8 bits. But for example what happens if I sample over 64 bits? Why 8 bits were chosen initially? For simplicity of types (`u8`) with the rest of the transcript things (and not having to deal with an exotic u64)?

Does this choice affect soundness?

### DIFF
--- a/src/whir/utils.rs
+++ b/src/whir/utils.rs
@@ -1,6 +1,7 @@
 use itertools::Itertools;
 use p3_challenger::{FieldChallenger, GrindingChallenger};
 use p3_field::{ExtensionField, Field, TwoAdicField};
+use p3_util::log2_ceil_usize;
 use tracing::instrument;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


WizardOfMenlo:
For context, in this step the verifier needs to sample a number of challenges, each one of which correspond to an in-domain sample. The total amount of bits to squeeze is then num_challenges * log (domain_size/2^folding_factor).
For example, in WHIR when doing 100 challenges over a domain of size 2^30 and with folding factor k = 4 the verifier needs to squeeze 100 * 26 ~ 2600 bits ~ 325 B.

tcoratger:
> For context, in this step the verifier needs to sample a number of challenges, each one of which correspond to an in-domain sample. The total amount of bits to squeeze is then num_challenges * log (domain_size/2^folding_factor). For example, in WHIR when doing 100 challenges over a domain of size 2^30 and with folding factor k = 4 the verifier needs to squeeze 100 * 26 ~ 2600 bits ~ 325 B.
> 
> If you squeeze them as bits or bytes or bananas (as long as there is a distribuiton preserving map from those to uniformly sample queries) everything will work out.

Ok perfect thanks for the details @WizardOfMenlo then can you check that my last commit works? I think it simplifies things

WizardOfMenlo:
At an high level it seems alright, it might be a bit more efficient in some cases to sample the whole bit vector at once and then parse the individual queries from it (as that reduces the sponge invocations). For example, if the domain size was 9 bits and you have to squeeze 4 queries naively you would do 8 sponge squeezes while you could just squeeze 5 times once and be done with it.

tcoratger:
> At an high level it seems alright, it might be a bit more efficient in some cases to sample the whole bit vector at once and then parse the individual queries from it (as that reduces the sponge invocations). For example, if the domain size was 9 bits and you have to squeeze 4 queries naively you would do 8 sponge squeezes while you could just squeeze 5 times once and be done with it.

Thanks will merge this first so we have something correct and track the performance thing with an issue :)

PR #197, opened 2025-06-30
Transform prover data in Witness to use a smart pointer

0xKarl98:
I have made changes according to this proposal

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,8 +10,8 @@ rust.missing_debug_implementations = "warn"
 rust.unreachable_pub = "warn"
 rust.unused_must_use = "deny"
 rust.rust_2018_idioms = { level = "deny", priority = -1 }
-rustdoc.all = "warn"
 rust.dead_code = "allow"
+rustdoc.all = "warn"
 
 [lints.clippy]
 # all lints that are on by default (correctness, suspicious, style, complexity, perf)
@@ -64,6 +64,7 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features = 
 ] }
 rand = "0.9"
 clap = { version = "4.4.17", features = ["derive"] }
+p3-keccak = "0.3.0"
 
 [dev-dependencies]
 criterion = "0.6"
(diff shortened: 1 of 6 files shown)


src/whir/prover/round.rs
```diff
@@ -50,7 +52,7 @@ where
 
     /// Merkle commitment prover data for the **base field** polynomial from the first round.
     /// This is used to open values at queried locations.
-    pub(crate) commitment_merkle_prover_data: MerkleTree<F, W, M, DIGEST_ELEMS>,
+    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, F, M, DIGEST_ELEMS>>,
```
tcoratger, line 55:
```suggestion
    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, W, M, DIGEST_ELEMS>>,
```

Cargo.toml
```diff
@@ -64,6 +64,7 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features =
 ] }
 rand = "0.9"
 clap = { version = "4.4.17", features = ["derive"] }
+p3-keccak = "0.3.0"
```
tcoratger, line 67: Is there a particular reason why we need this now? This should not be used in the code after recent updates I think since everything should be Poseidon now no?

tcoratger (commented):
Small nits and then we are good to merge.

src/whir/prover/round.rs
0xKarl98: I will reapply your suggestion

src/whir/mod.rs
```diff
@@ -27,7 +27,7 @@ pub mod utils;
 pub mod verifier;
 
 type F = BabyBear;
-type EF = BinomialExtensionField<F, 4>;
+type EF4 = BinomialExtensionField<F, 4>;
```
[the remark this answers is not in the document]
tcoratger: In the other places this looks ok to keep `EF4` because these are specific unit tests and changing the field will probably not change the test at all but here it's a bit different because this is the end to end test. Maybe one day, we will want to test Goldilocks extension field of degree 2 or any other field.

PR #198, opened 2025-06-30
Transform prover data in Witness to use a smart pointer 

### LINKED ISSUE #176: Transform prover data in `Witness` to use a smart pointer
### Problem

Currently, the `Witness` struct owns the Merkle tree prover data directly:

https://github.com/tcoratger/whir-p3/blob/54a320eb95e0acdca14aefa116aa0ab3033d51f4/src/whir/committer/mod.rs#L13-L32

This design can lead to unnecessary cloning or awkward lifetime management when passing the `Witness` across layers of the proof system.

A `TODO` already hints at the issue (made by @han0110)

```rust
// TODO(whir-p3): Use reference to merkle tree in `Witness` to avoid cloning or ownership taking.
```

### Proposal

Transform `prover_data` to a smart pointer:

```rust
use std::sync::Arc;
pub prover_data: Arc<MerkleTree<F, W, M, DIGEST_ELEMS>>,
```

### Benefits

* Avoids ownership and cloning issues.
* Enables shared access without leaks (`Box::leak` workaround not needed).
* Retains Rust's safety guarantees.
* Cleaner and more ergonomic than using `RefCell` or manual memory hacks.

### Alternatives Considered

* `Box`: Works but requires split return (`(Witness, Box<MerkleTree>)`).
* `RefCell`: Enables late `take()` but introduces runtime borrow checks.
* `Arc`: Cleanest and most ergonomic, especially if cloning is cheap and needed in multiple places.

@han0110 What do you think about using `Arc` here?

@0xKarl98 Maybe you could be interested by this one?

0xKarl98:
This closes #176 and applied all suggestions

### DIFF
--- a/src/whir/prover/round.rs
+++ b/src/whir/prover/round.rs
@@ -52,7 +52,7 @@ where
 
     /// Merkle commitment prover data for the **base field** polynomial from the first round.
     /// This is used to open values at queried locations.
-    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, F, M, DIGEST_ELEMS>>,
+    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, W, M, DIGEST_ELEMS>>,
 
     /// Merkle commitment prover data for the **extension field** polynomials (folded rounds).
     /// Present only after the first round.
--- a/src/whir/prover/round.rs
+++ b/src/whir/prover/round.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use p3_challenger::{FieldChallenger, GrindingChallenger};
-use p3_field::{ExtensionField, Field, TwoAdicField};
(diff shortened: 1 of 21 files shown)


src/whir/prover/round.rs
```diff
@@ -50,7 +52,7 @@ where
 
     /// Merkle commitment prover data for the **base field** polynomial from the first round.
     /// This is used to open values at queried locations.
-    pub(crate) commitment_merkle_prover_data: MerkleTree<F, W, M, DIGEST_ELEMS>,
+    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, F, M, DIGEST_ELEMS>>,
```
tcoratger, line 55:
```suggestion
    pub(crate) commitment_merkle_prover_data: Arc<MerkleTree<F, W, M, DIGEST_ELEMS>>,
```

tcoratger (approved):

PR #203, opened 2025-07-05
prover: activate packing

tcoratger:


TomWambsgans:
Nice! Remark: maybe we could avoid using packing for the verifier ? For a maximum of readability

tcoratger:
> Nice! Remark: maybe we could avoid using packing for the verifier ? For a maximum of readability

Yes you are right, I hesitated but you are right I will revert this part and in any case I don't have the impression that it changes much to verify it from my experiments.

PR #205, opened 2025-07-07
Apply #192 proposal improvements

### LINKED ISSUE #192: Optimize stir query sampling
Related: https://github.com/tcoratger/whir-p3/pull/191#issuecomment-3009324360

In the current implementation of `get_challenge_stir_queries`, we sample each query index independently by calling `challenger.sample_bits(domain_size_bits)` for every query.

https://github.com/tcoratger/whir-p3/blob/78bbf06cd32f9ba7ab2298a3506cc8bf0996a976/src/whir/utils.rs#L24-L62

While this is functionally correct and clean, it may not be the most efficient approach — especially when the number of bits per query is small and the number of queries is large.

### Observation

Each call to `sample_bits(k)` typically results in one sponge invocation. So sampling `n` queries costs `n` squeezes, even if the total number of bits is small enough to be squeezed all at once.

> For example, if `domain_size_bits = 9` and `num_queries = 4`, then:
>
> * We need `9 × 4 = 36` bits total.
> * This could be squeezed from the transcript in just 5 bytes.
> * But calling `sample_bits(9)` four times results in **4 separate sponge squeezes**.

This is a missed optimization opportunity.

### Suggested Improvement

Instead of calling `sample_bits(domain_size_bits)` `num_queries` times, we can:

1. Compute the **total number of bits** needed:

   ```rust
   total_bits = num_queries * domain_size_bits;
   ```
2. Squeeze the full bit vector **once** (e.g. as bytes or u64 chunks).
3. Parse each query from that bitstream directly.

This would **reduce sponge invocations**, improving performance without changing behavior or affecting soundness.

### Action Item

- Try to optimize the function
- Add a criterion benchmark file to validate the perf improvement properly under various context (small, medium, large number of queries)

0xKarl98:
This closes #192 

For optimizations benchmark file , please check this out :
https://hackmd.io/2gywYFPaTlC3TCrwGNDqJA

### DIFF
--- a/benches/challenge_queries.rs
+++ b/benches/challenge_queries.rs
@@ -2,8 +2,7 @@ use std::hint::black_box;
 
 use criterion::{Criterion, criterion_group, criterion_main};
 use p3_baby_bear::BabyBear;
-use whir_p3::fiat_shamir::ChallengSampler;
-use whir_p3::whir::utils::get_challenge_stir_queries;
+use whir_p3::{fiat_shamir::ChallengSampler, whir::utils::get_challenge_stir_queries};
 
 type F = BabyBear;
 


PERFORMANCE_IMPROVEMENTS.md
```diff
@@ -0,0 +1,148 @@
+# Performance Optimization Summary
```
tcoratger, line 1: Can we remove this file and just put all this HackMD in the PR description only. I feel like this is a good idea to have markdown documentation for the whole code but probably we need to start with more fundamental elements than this (and this is something that we can do later). But all this recap is very good for further checking by developers interested in the tricks so I don't want to lose it. Please put it in the PR description so that everyone can have access when needed.

improved_challenge_queries.rs
```diff
@@ -0,0 +1,91 @@
+use crate::utils::log2_ceil_usize;
```
tcoratger, line 1: Why do we need all this code in the root of the repo? `get_challenge_stir_queries` function already exists in the utils no?

src/parameters/mod.rs
```diff
@@ -167,6 +167,72 @@ impl FoldingFactor {
...
+        if self.max_bits_per_call == 0 {
+            return Err("max_bits_per_call must be greater than 0");
+        }
+        if self.max_bits_per_call > 63 {
+            return Err("max_bits_per_call must not exceed 63 to prevent overflow");
+        }
+        Ok(())
+    }
+}
+
```
tcoratger, line 235: I don't think it's the best idea to put all of this as tweakable parameters because it can confuse the user, who isn't necessarily aware of the precise meaning of each parameter.
- Either I will be confused and won't really know what to do with these parameters.
- Or I will fill in inappropriate parameters, and the computation could be worse.

Can we have these two parameters and constants in the util file?

src/whir/utils.rs
```diff
@@ -21,44 +25,183 @@ pub const fn workload_size<T: Sized>() -> usize {
...
-/// A sorted and deduplicated list of random query indices in the folded domain.
-pub fn get_challenge_stir_queries<Chal: ChallengSampler<EF>, F, EF>(
+/// Sorted and deduplicated list of query indices in the folded domain
+pub fn get_challenge_stir_queries<Challenger, F>(
     domain_size: usize,
     folding_factor: usize,
     num_queries: usize,
-    prover_state: &mut Chal,
+    challenger: &mut Challenger,
+    config: Option<&ChallengeQueryConfig>,
```
tcoratger, line 142: Here I think that we prefer working with the state instead of challenger directly to avoid confusion, cf https://github.com/tcoratger/whir-p3/pull/202

src/whir/utils.rs
```diff
@@ -21,44 +25,183 @@ pub const fn workload_size<T: Sized>() -> usize {
...
+        let mut queries = Vec::with_capacity(num_queries);
+
+        for _ in 0..num_queries {
+            let query = reader.read_bits(domain_size_bits) % folded_domain_size;
+            queries.push(query);
+        }
+
+        // Sort and remove duplicates
+        queries.sort_unstable();
+        queries.dedup();
```
tcoratger, line 202: do we really need all this complexity here? I fill like this is overkill, can we just have something similar to this?

https://github.com/WizardOfMenlo/whir/blob/88cbde0bff773104af4a09c639a32b6eff339e82/src/whir/utils.rs#L57-L85

with just the chunk system in a very simple way? I would love to avoid additional complexity as much as possible

0xKarl98:
@tcoratger Please check it out

0xKarl98:
@tcoratger  Do you think if we are good to merge this or need furthur modification ?

tcoratger:
> @tcoratger Do you think if we are good to merge this or need furthur modification ?

Sorry for the delay, I'll review this today or tomorrow :)

src/parameters/mod.rs
```diff
@@ -167,6 +167,72 @@ impl FoldingFactor {
...
+        }
+        if self.max_bits_per_call == 0 {
+            return Err("max_bits_per_call must be greater than 0");
+        }
+        if self.max_bits_per_call > 63 {
+            return Err("max_bits_per_call must not exceed 63 to prevent overflow");
+        }
+        Ok(())
+    }
+}
```
tcoratger, line 234: We can delete all this no?

tcoratger:
Have you re-run the benchmarks to see the improvements compared to the original version?

0xKarl98:
I have re-run the benmarks and performance has regressed , gonna think about this again

tcoratger:
I guess this one is outdated, don't hesitate to open a new one if you feel something can be improved :)

PR #207, opened 2025-07-09
`add_new_equality`: `eval_eq` instead of chunked approach

tcoratger:


tcoratger:
Closing in favor of https://github.com/tcoratger/whir-p3/pull/209

PR #209, opened 2025-07-11
Use add_slices in eval_eq

SyxtonPrime:
My computer is being a little dodgy so I'm not sure how much this changes the timings. (Can someone check?)

In principle, using `add_slices` should be quite a bit faster as it makes use of packing internally. In practice though I suspect this isn't really the part of the function which is particularly slow.

### DIFF
--- a/src/sumcheck/sumcheck_single.rs
+++ b/src/sumcheck/sumcheck_single.rs
@@ -477,26 +477,66 @@ where
             const PARALLEL_THRESHOLD: usize = 4096;
 
             match &self.evaluation_of_p {
-                EvaluationStorage::Base(evals_f) => (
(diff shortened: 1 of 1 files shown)


tcoratger:
So, it's faster than what's currently on main for sure. I ran multiple run to be sure and I just take the largest call during the first round by running the main program to stress test: `RUSTFLAGS='-C target-cpu=native' cargo run --release`
- On current main (commit 8084da1): `add_new_equality [ 351ms | 20.98% ] num_points: 82`
- On #207: `add_new_equality [ 283ms | 23.78% ] num_points: 82`
- On this PR: `add_new_equality [ 233ms | 0.00% / 15.24% ] num_points: 82`

tcoratger (approved):

PR #223, opened 2025-07-23
fix grinding (was done after query, instead of before)

TomWambsgans:
As described in section 6.3 of the [EthStark paper](https://eprint.iacr.org/2021/582.pdf), grinding should be done "before" queries. Because a typical attack against FRI/WHIR is to get "lucky" indexes, after having sent a polynomial which does not correspond to the folding of the previously committed (via merkle root of evaluations) polynomial. So we want the grinding work to appear during each iteration of the search for these "lucky" indexes, not after, once they are found.


TomWambsgans:
🔥

PR #224, opened 2025-07-24
Preparation for an optimization: remove embedding overhead

TomWambsgans:
Preparation for an optimization: there is currently an embedding overhead, in `add_new_equality`, most of the points come from stir queries and are in fact in F, not in EF. This PR does not fully fixes the issue, but does 50% of the job. The remaining part is to write a function similar to `utils::eval_eq`, but with `eval` in F, not in EF.


src/whir/prover/mod.rs
```diff
@@ -309,10 +310,7 @@ where
                 }
 
                 // Evaluate answers in the folding randomness.
-                let mut stir_evaluations = ood_answers;
-                // Exactly one growth
-                stir_evaluations.reserve_exact(answers.len());
-
+                let mut stir_evaluations = vec![];
```
tcoratger, line 313:
```suggestion
                let mut stir_evaluations = Vec::with_capacity(answers.len());
```

src/whir/prover/mod.rs
```diff
@@ -344,10 +342,7 @@ where
...
```
tcoratger, line 345:
```suggestion
                let mut stir_evaluations = Vec::with_capacity(answers.len());
```

src/whir/prover/mod.rs
```diff
@@ -361,15 +356,29 @@ where
 
         // Randomness for combination
         let combination_randomness_gen: EF = prover_state.sample();
-        let combination_randomness: Vec<_> = combination_randomness_gen
+        let ood_combination_randomness: Vec<_> = combination_randomness_gen
+            .powers()
+            .take(ood_challenges.len())
+            .collect();
```
tcoratger, line 362:
```suggestion
        let ood_combination_randomness = combination_randomness_gen
            .powers()
            .collect_n(ood_challenges.len());
```

src/whir/prover/mod.rs
```diff
@@ -361,15 +356,29 @@ where
...
+        round_state.sumcheck_prover.add_new_equality(
+            &ood_challenges,
+            &ood_answers,
+            &ood_combination_randomness,
+        );
+        let stir_combination_randomness: Vec<_> = combination_randomness_gen
             .powers()
+            .skip(ood_challenges.len())
             .take(stir_challenges.len())
             .collect();
```
tcoratger, line 372:
```suggestion
        let stir_combination_randomness = combination_randomness_gen
            .powers()
            .skip(ood_challenges.len())
            .collect_n(stir_challenges.len());
```

src/poly/multilinear.rs
```diff
@@ -185,6 +185,11 @@ where
 
         acc
     }
+
+    #[must_use]
+    pub fn embed<EF: ExtensionField<F>>(&self) -> MultilinearPoint<EF> {
+        MultilinearPoint(self.0.iter().map(|&x| EF::from(x)).collect())
+    }
```
tcoratger, line 192: Can we have a 1-2 line header doc to explain what this function does and why? :)

src/whir/parameters.rs
```diff
@@ -259,6 +249,10 @@ where
         }
     }
 
+    pub const fn starting_domain_size(&self) -> usize {
+        1 << (self.mv_parameters.num_variables + self.starting_log_inv_rate)
+    }
```
tcoratger, line 254:
```suggestion
/// Returns the size of the initial evaluation domain.
///
/// This is the size of the domain used to evaluate the original multilinear polynomial
/// before any folding or reduction steps are applied in the WHIR protocol.
///
/// It is computed as:
/// 
/// \begin{equation}
/// 2^{\text{num\_variables} + \text{starting\_log\_inv\_rate}}
/// \end{equation}
///
/// - `num_variables` is the number of variables in the original multivariate polynomial.
/// - `starting_log_inv_rate` is the initial inverse rate of the Reed–Solomon code,
///    controlling redundancy relative to the degree.
///
/// # Returns
/// A power-of-two value representing the number of evaluation points in the starting domain.
    pub const fn starting_domain_size(&self) -> usize {
        1 << (self.mv_parameters.num_variables + self.starting_log_inv_rate)
    }
```

tcoratger (commented):
Looks good to me, just minor comments

PR #233, opened 2025-07-28
Angus/add new equality base

### LINKED ISSUE #227: embedding overhead in `add_new_equality`
In most of the cases, `add_new_equality` operates in reality over the base field, but currently we convert it to the extension field for simplicity, see [here](https://github.com/tcoratger/whir-p3/blob/647befe6f813cf5f2ece7de4aad6135ee0f75a26/src/whir/prover/mod.rs#L377). As a result there is room for optimization

SyxtonPrime:
Another take on #232  which should ideally close #227.

Currently quite a bit of copy/pasted code so I want to cut down on that before merging.

Seems to give a 20-30% speed up (looking at the time taken for the second ` accumulate_weight_buffer` when running cargo --run --release). Want to try out a couple of different approaches to.


SyxtonPrime:
Yep no worries, I should be able to get it ready to merge in the next day or two. Just think I can reduce the amount of copy and paste.

SyxtonPrime:
Yep, I think it's good to go.

PR #241, opened 2025-08-10
Derive first coeff of sumcheck poly

kilic:
Prover can send `c0` and `c2` of sumcheck polynomial and verifier can recover `c1` by doing cheap `current_sum - c0`. So for each sumcheck round, this PR saves 1 ext field element from proof data


tcoratger (approved):
Looks good to me, what do you think @TomWambsgans ? Normally, this shouldn't make it any more difficult when using univariate skip, as it only applies to classical rounds.

TomWambsgans:
Looks good to me as well

Issue #234, opened 2025-07-31
Sumcheck optimization


WizardOfMenlo:
Might be nice to change the evaluations points of the sumcheck polynomial to speed up proving. 
I think at the moment we use 0,1,2. Instead I think 0,1,1/2 will be kind of nicer.
The reason is that `eq(1/2, 0) = 1/2 * 0+ (1 - 1/2)  * 1 = 1/2` and `eq(1/2, 1) = 1/2 * 1 + 0 * 1/2 = 1/2`.
So, since this is symmetric, to compute `h(1/2) = sum_b p(1/2, b) * q(1/2, b) = sum_b (eq(1/2, 0) * p(0, b) + eq(1/2, 1) * p(1,b)) * (eq(1/2, 0) * q(0, b) + eq(1/2, 1) * q(1,b)) = 1/4 * sum_b (p(0, b) + p(1, b)) * (q(0, b) + q(1,b))`.
i.e. we only need 1 multiplication per hypercube element.

PR #254, opened 2025-08-22
sumcheck: change evaluation points

tcoratger:
Related: https://github.com/tcoratger/whir-p3/issues/234#issue-3281741420


src/sumcheck/sumcheck_polynomial.rs
```diff
@@ -121,11 +128,29 @@ where
...
+            self.n_variables, 1,
+            "Only implemented for univariate polynomials"
+        );
+
+        let x = point[0];
+        let h0 = self.evaluations[0];
+        let h1 = self.evaluations[1];
+        let h_half = self.evaluations[2];
+
+        let one = F::ONE;
```
diegokingston, line 141: As an alternative,
```
        let one = F::ONE;
        let two_x = x.double();

        // L_0(x) = 2(x-1)(x-1/2) apply distributive
        let l0 = (x - one) * (two_x - one);
        // L_1(x) = 2x(x-1/2) apply distributive
        let l1 =  x * (two_x - one);
        // L_2(x) = -4x(x-1)
        let l2 =  two_x.double() * (one - x);
```
This should save some computations. Another alternative could be:
c0 = h0;
c1 = 2*(h0 + h1 - 2h_half);
c2 = h1 - (c1 + c0);
And then the evaluation is (c2*x+c1)*x+c0, which solves the evaluation in two field multiplications, instead of the six we have for Lagrange polynomials

tcoratger: Thanks for the suggestion here, just applied it. 

I'm a little surprised because now the `evaluate_on_standard_domain` function only has 3 multiplications, whereas it had more before:
- 3 due to this one `eval * point.eq_poly3(i)` in the loop
- Some other ones in `eq_poly3`

But despite this, it seems that the benchmarks aren't any faster than before. Here's what I get on my computer when I run the sumcheck benchmark with the new version compared to the new one.

```rust
Gnuplot not found, using plotters backend
SumcheckProver/Classic/16
                        time:   [2.0243 ms 2.0399 ms 2.0529 ms]
                        change: [+0.7800% +1.7474% +2.6323%] (p = 0.00 < 0.05)
                        Change within noise threshold.
SumcheckProver/Classic/18
                        time:   [5.1560 ms 5.2144 ms 5.3152 ms]
                        change: [+0.2751% +3.0167% +5.8184%] (p = 0.05 > 0.05)
                        No change in performance detected.
SumcheckProver/Classic/20
                        time:   [18.042 ms 18.486 ms 19.195 ms]
                        change: [+14.570% +19.105% +24.290%] (p = 0.00 < 0.05)
                        Performance has regressed.
SumcheckProver/Classic/22
                        time:   [45.446 ms 47.558 ms 52.208 ms]
                        change: [+0.6889% +9.9423% +20.252%] (p = 0.05 > 0.05)
                        No change in performance detected.
Benchmarking SumcheckProver/Classic/24: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 9.2s or enable flat sampling.
SumcheckProver/Classic/24
                        time:   [164.46 ms 169.18 ms 174.53 ms]
                        change: [−0.3630% +3.9793% +9.3162%] (p = 0.14 > 0.05)
                        No change in performance detected.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
  ```

diegokingston: In the case of quadratic sumcheck, I think we don't need to evaluate x at 1/2 or 0, since we know the form of the polynomial. For t(r1,r2,x,b) = p(r1,r2,x,b)*q(r1,r2,x,b) we know that p(r1,r2,x,b) = p(r1,r2,0,b)(1-x) + p(r1,r2,1,b)x and similarly for q(r1,r2,x,b) = q(r1,r2,0,b)(1-x) + q(r1,r2,1,b)x, so we can write p(r1,r2,x,b) = p(r1,r2,0,b) + (p(r1,r2,1,b) - p(r1,r2,0,b))x and so we can work out the product polynomial a0 + a1 x + a2 x^2. Both a0 and a2 can be computed from p(r1,r2,0,b)q(r1,r2,0,b) and (p(r1,r2,1,b)-p(r1,r2,0,b))*(q(r1,r2,1,b)-q(r1,r2,0,b)). a1 can be computed from the previous round and then we don't need any further interpolation. 
As for specific performance loss, I guess that doing multiplication by 1/2 is a bit more expensive (using Montgomery) than multiplying by two and subtracting (which are essentially additions).

tcoratger: After thinking about this and as the integration with Plonky3 with @SyxtonPrime progressed, I thought about this: https://github.com/tcoratger/whir-p3/pull/265

The benchmarks included in the PR for the `evaluate_on_standard_domain` function (we completely eliminated the `eval_eq3` function) show amazing results. What do you think about this?

diegokingston: Yes, sounds good. This PR shows that evaluating at 1/2 does not offer an advantage (we did a separate implementation to check it), so I see no need to continue on this line.

PR #289, opened 2025-09-15
Adding methods to constraint

SyxtonPrime:
The goal of this PR is to add clarity around the purpose of the constraint struct.

Working through this I do wonder if there is a better way to encapsulate this idea (essentially get rid of `Constraint` and focus on `Statement` as the core type) I might try that in a separate PR. Currently, outside of `verify`, every other method on `Constraint` is either a constructor or a destructor which suggests that this might not be the right encapsulation.

Regardless, the main changes in this PR are:
- Remove `defer_evaluation` and everything referring to it.
- Add a `new` method and methods to retrieve the point and the claimed evaluation.
- Remove `pub` from the point and claimed evaluation. (These should not be changed after the constraint is initialized so should probably not be pub)
- Move `combine_constraints` which has been copied in `2` places into the constraint file.

### DIFF
--- a/src/sumcheck/tests.rs
+++ b/src/sumcheck/tests.rs
@@ -16,7 +16,7 @@ use crate::{
     },
     poly::{evals::EvaluationsList, multilinear::MultilinearPoint},
     whir::{
-        statement::{Statement, constraint::Constraint},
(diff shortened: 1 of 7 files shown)


tcoratger (approved):
Just fmt and then we are good to merge.

PR #290, opened 2025-09-15
Remove Constraint

SyxtonPrime:
Currently the only method for constraint which is not a constructor or destructor is the the `verify` method.

Due to this `Constraint` seems to be a struct which is not really encapsulating anything particularly important. Hence this PR looks to remove it and absorb its functionality into `Statement`.

Full set of changes are as follows:

**Constraint:**
- Remove the Constraint file and all code.

**Statement:**
- Now holds 2 vectors, one of points and one of evaluations. All fields are private.
- Rename `new` to `initialize`.
- Make a new `new` method which takes a vector of points and evaluations and does some consistency checks.
- Make a new methods: `is_empty, iter, len` which do basically what you'd expect. (iter returns an iterator over pairs).
- New method `add_unevaluated_constraint` which lets you add a new constraint with a point and a polynomial. E.g. Handles computing the evaluation for you.
- New method `verify` which verifies that a given polynomial satisfies all constraints in the statement.
- New method: concatenate which lets us concatenate two statements together.
- New method: get_points which takes ownership and returns the points vector. Don't love this solution but I think it's better than making `points` public. This is mostly unused atleast.
- remove: `add_constraint_in_front` as it's unused.
- Currently left `add_constraints_in_front` in but I think we could probably remove it as concatenate should be able to be used instead. Leave that for later though.
- New method: `combine_evals` which does a linear combination using powers of a random variable to accumulate the evaluations.

TODO: We should probably move statement to its own file (as opposed to sitting in mod.rs) now that it's gotten pretty big. Will leave that for a future PR though.


src/sumcheck/tests.rs
```diff
@@ -271,7 +265,7 @@ fn eval_constraints_poly<F, EF>(
     num_vars: usize,
     k_skip: usize,
     folding_factors: &[usize],
-    constraints: &[Vec<Constraint<EF>>],
+    constraints: &[Statement<EF>],
```
tcoratger, line 268:
```suggestion
    statements: &[Statement<EF>],
```

src/whir/statement/mod.rs
```diff
@@ -64,14 +77,49 @@ impl<F: Field> Statement<F> {
         self.num_variables
     }
 
+    /// Returns true if the statement contains no constraints.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        debug_assert_eq!(
+            self.evaluation_points.is_empty(),
+            self.expected_evaluations.is_empty()
+        );
```
tcoratger, line 86: I think that if you use `debug_assert` we could make this function const

src/whir/statement/mod.rs
```diff
@@ -64,14 +77,49 @@ impl<F: Field> Statement<F> {
...
+            .zip(self.expected_evaluations.iter())
+    }
+
+    /// Returns the number of constraints in the statement.
+    #[must_use]
+    pub fn len(&self) -> usize {
+        debug_assert_eq!(
+            self.evaluation_points.len(),
+            self.expected_evaluations.len()
+        );
```
tcoratger, line 103: With `debug_assert` I think that we could make this function constant.

tcoratger (approved):
This makes a lot of sense, just minor nits and small header docs for some methods that miss doc and then we can merge. After this, I can close the Plonky3 Constraint PR and replace it by a Statement PR (to integrate the statement mod file).

src/whir/statement/mod.rs
```diff
@@ -132,32 +187,57 @@ impl<F: Field> Statement<F> {
...
+    /// - `alpha`: A random extension field element used to weight the constraints.
+    ///
+    /// # Returns
+    /// A `Vec<EF>` containing the powers of `alpha` used to combine each constraint.
+    pub fn combine_evals(&self, claimed_eval: &mut F, gamma: F) -> Vec<F> {
+        combine_evals(claimed_eval, &self.expected_evaluations, gamma)
+    }
+}
+
+pub fn combine_evals<F: Field>(claimed_eval: &mut F, evals: &[F], gamma: F) -> Vec<F> {
```
tcoratger, line 233: as we have the `combine_evals` in the statement implementation, do we need to have this public (don't know if this is used elsewhere)? If not maybe we can make this function private to avoid confusion for the user.

tcoratger (approved):
Two small nits and then we can merge

src/whir/statement/mod.rs
SyxtonPrime: Actually I think I can get rid of it.

PR #295, opened 2025-09-16
Faster compression of eqs

kilic:
This PR speeds up `Statement::combine` function that compresses multiple eqs with a randomness. Compression itself has more than 2x speed and `prove` function reduces from ~900ms to ~600ms when I run `RUSTFLAGS='-C target-cpu=native' cargo run --release `

### DIFF
--- a/src/whir/constraints/statement.rs
+++ b/src/whir/constraints/statement.rs
@@ -1,8 +1,5 @@
 use p3_field::{ExtensionField, Field};
-use rayon::{
-    iter::{IndexedParallelIterator, ParallelIterator},
-    slice::{ParallelSlice, ParallelSliceMut},
-};
+use p3_maybe_rayon::prelude::*;
 use tracing::instrument;
 
 use crate::poly::{evals::EvaluationsList, multilinear::MultilinearPoint};
--- a/src/whir/constraints/statement.rs
+++ b/src/whir/constraints/statement.rs
@@ -159,39 +159,54 @@ impl<F: Field> Statement<F> {
             );
         }
 
+        // Number of constraints (columns).
         let n = self.len();
+        // Number of variables (bits).
         let k = self.num_variables();
 
-        // Allocate space for all eqs at once
+        // Allocate a flat buffer for the (2^k × n) matrix of weighted equality rows.
+        //
+        // Safety: we fill every entry before reading it.
         let mut eq: Vec<F> = unsafe { crate::utils::uninitialized_vec((1 << k) * n) };
 
-        // Prepare RLC coeffs
+        // Precompute γ^i for i = 0..n-1 (random linear-combination weights).
         let challenges = challenge.powers().take(n).collect();
 
-        // Initialize the first row with randomness factors
+        // Initialize row 0 with [γ^0, γ^1, …, γ^{n-1}].
         eq.iter_mut()
             .zip(challenges.iter())
             .for_each(|(e, alpha)| *e = *alpha);
 
-        // Build eqs independently
+        // Expand row 0 into 2^k rows with a simple two-branch split per bit.
+        // For bit i (from most significant to least):
+        //   high  = low * z
+        //   low   = low - high  = low * (1 - z)
+        // Here z ∈ {0,1} is the i-th coordinate of the constraint point.
         for i in 0..k {
+            // Split the first 2^i rows into low and high halves in place.
             let (lo, hi) = eq.split_at_mut((1 << i) * n);
+            // Work in parallel over row pairs. Each pair has n columns.
             lo.par_chunks_mut(n)
                 .zip(hi.par_chunks_mut(n))
                 .for_each(|(lo, hi)| {
+                    // For each column j: read its constraint point, update the pair.
                     self.points
                         .iter()
                         .zip(lo.iter_mut())
                         .zip(hi.iter_mut())
-                        .for_each(|((point, a0), a1)| {
-                            *a1 = *a0 * point[k - i - 1]; // reversed point order
-                            *a0 -= *a1;
+                        .for_each(|((point, lo), hi)| {
+                            // Take the current bit from the end (most significant first)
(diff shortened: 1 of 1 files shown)


src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
-
-        // Process the remaining constraints.
-        for (point, &eval) in rest_points.iter().zip(rest_evals) {
-            // Update γ to γ^i for this constraint.
-            gamma *= challenge;
+        let n = self.len();
+        let k = self.num_variables();
+
+        // Allocate space for all eqs at once
+        let mut eq: Vec<F> = unsafe { crate::utils::uninitialized_vec((1 << k) * n) };
```
tcoratger, line 166:
```suggestion
        // Allocate a flat buffer for the (2^k × n) matrix of weighted equality rows.
        //
        // Safety: we fill every entry before reading it.
        let mut eq: Vec<F> = unsafe { crate::utils::uninitialized_vec((1 << k) * n) };
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+
+        // Prepare RLC coeffs
```
tcoratger, line 168:
```suggestion
        // Precompute γ^i for i = 0..n-1 (random linear-combination weights).
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+        let challenges = challenge.powers().take(n).collect();
+
+        // Initialize the first row with randomness factors
```
tcoratger, line 171:
```suggestion
        // Initialize row 0 with [γ^0, γ^1, …, γ^{n-1}].
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
-        // Initialize the combined expected sum with the first term: s_1 * γ^0 = s_1.
-        let mut gamma = F::ONE;
-        let mut sum = *first_eval;
-
-        // Process the remaining constraints.
-        for (point, &eval) in rest_points.iter().zip(rest_evals) {
-            // Update γ to γ^i for this constraint.
-            gamma *= challenge;
+        let n = self.len();
+        let k = self.num_variables();
```
tcoratger, line 163:
```suggestion
        // Number of constraints (columns).
        let n = self.len();
        // Number of variables (bits).
        let k = self.num_variables();
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+
+        // Prepare RLC coeffs
+        let challenges = challenge.powers().take(n).collect();
+
+        // Initialize the first row with randomness factors
+        eq.iter_mut()
+            .zip(challenges.iter())
+            .for_each(|(e, alpha)| *e = *alpha);
+
+        // Build eqs independently
```
tcoratger, line 176:
```suggestion
    // Expand row 0 into 2^k rows with a simple two-branch split per bit.
    // For bit i (from most significant to least):
    //   high  = low * z
    //   low   = low - high  = low * (1 - z)
    // Here z ∈ {0,1} is the i-th coordinate of the constraint point.
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+        for i in 0..k {
+            let (lo, hi) = eq.split_at_mut((1 << i) * n);
```
tcoratger, line 178:
```suggestion
            // Split the first 2^i rows into low and high halves in place.
            let (lo, hi) = eq.split_at_mut((1 << i) * n);
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+            lo.par_chunks_mut(n)
```
tcoratger, line 179:
```suggestion
            // Work in parallel over row pairs. Each pair has n columns.
            lo.par_chunks_mut(n)
```

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,54 @@ impl<F: Field> Statement<F> {
...
+                .zip(hi.par_chunks_mut(n))
+                .for_each(|(lo, hi)| {
+                    self.points
```
tcoratger, line 182:
```suggestion
                    // For each column j: read its constraint point, update the pair.
                    self.points
```

tcoratger (commented):
Thanks a lot for this! Just reworked a bit the doc, but the code is amazing! Nice trick!

tcoratger:
Just an idea (@SyxtonPrime should confirm or not), as we are operating over a matrix like object, maybe we could use a `DenseMatrix` directly from here https://github.com/Plonky3/Plonky3/blob/main/matrix/src/dense.rs

src/whir/constraints/statement.rs
```diff
@@ -158,36 +159,69 @@ impl<F: Field> Statement<F> {
...
+            .collect::<Vec<_>>();
 
-            // Add this constraint's contribution to the combined expected sum:
-            sum += eval * gamma;
-        }
+        // Horner sums
+        let sum = self
+            .evaluations
+            .iter()
+            .rfold(F::ZERO, |acc, coeff| acc * challenge + *coeff);
```
SyxtonPrime, line 219: Given we have already computed the vector `challenges`, its a little simpler to just use that vector and do a dot product.
Should be the same speed though.

```suggestion
        let sum = dot_product(self.evaluations.iter().copied(), challenges.into_iter());
```

tcoratger (approved):
Thanks a lot for this!

PR #301, opened 2025-09-22
univariate skip: fix TODO

tcoratger:
@SyxtonPrime Should fix the TODO

### DIFF
--- a/src/sumcheck/sumcheck_single_skip.rs
+++ b/src/sumcheck/sumcheck_single_skip.rs
@@ -180,7 +180,7 @@ mod tests {
         let width = 1 << num_remaining_vars;
         let (poly, _, _) = compute_skipping_sumcheck_polynomial::<F, EF4>(
             evals.into_mat(width),
-            weights.into_mat(width)
+            weights.into_mat(width),
         );
 
         // ----------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


SyxtonPrime (approved):
lgtm

You should be able to remove a lot of the checks though.

src/sumcheck/sumcheck_single_skip.rs
```diff
@@ -42,9 +38,7 @@ use crate::poly::evals::EvaluationsList;
 #[must_use]
 #[instrument(skip_all)]
 pub(crate) fn compute_skipping_sumcheck_polynomial<F, EF>(
-    k: usize,
-    evals: &EvaluationsList<F>, // TODO: evals and weights should be given as a pair of matrices.
-    weights: &EvaluationsList<EF>,
+    (f_mat, weights_mat): (RowMajorMatrix<F>, RowMajorMatrix<EF>),
```
SyxtonPrime, line 41: Last quick comment. This should just accept 2 arguments right?

No reason to do a pair here.

tcoratger: Ah yeah as you suggested initially to put this in a pair of matrix, I found it elegant to translate the pair in a tuple but yeah just did 2 arguments, this is equivalent :)

PR #302, opened 2025-09-22
Commit to all three sumcheck polynomial coeffs

SyxtonPrime:
I think this might help us in tracking down the issues with the univariate sumcheck impl.

Can undo this PR at a later date.


tcoratger (approved):
Yes good thing to have

Issue #298, opened 2025-09-18
Thoughts on Eval Eq


SyxtonPrime:
Currently, `Eval_Eq` (and `Eval_Eq_Base`) is responsible for around `40 - 50%` of the total proof time. Primarily this time consumption occurs in a call to `combine` early on and the first call to `add_new_base_equality`. Hence this is an important place to look for time improvements.

One of the lessons from #295 is that there is definitely room to improve things here.

Unfortunately, this will likely require a reasonably substantial rework of `eval_eq` (and `eval_eq_base`) in order to get things working and also involve some substantial changes on this end so I thought it would be worth writing out my thoughts here.

Currently `Eval_Eq` essentially defines the map: $(\textbf{z}, \gamma) -> (\gamma \cdot eq(\textbf{z}, \textbf{x}))_{\textbf{x} \in B}$. Where B denotes the appropriate hypercube. This is a vector of length `2^n` and we additionally pass in an accumulation vector so instead of returning a vector, `eval_eq` simply adds the result into our accumulator.

This means that, for each $\textbf{x}$, the accumulator will end up holding $\sum \gamma_i \cdot eq(\textbf{z}_i, \textbf{x})$.

As each call of `Eval_Eq` costs roughly `2^n` muls and `2^{n + 1}` add/subs, if we want to accumulate `m` points, this will require roughly `m*2^n` multiplications and `m2^{n + 1}` add/subs. We also end up having to send a lot of work between cores.

I think we can do better on both fronts. To make things simple, consider the case where `n = 1` so eval_eq is simply the map $(z, \gamma, acc) -> acc_0 += \gamma (1 - z), acc[1] += \gamma z$. (Note that the `2^n/2^{n + 1}` above is really `2^n - 1` which makes a big difference in exactly this case).

We can see that we need `1` mul to get $\gamma z$, `1` sub to get $\gamma (1 - z) = \gamma - z$ and $2$ adds to add into the accumulator. Thus for $m$ points the cost here is `m` muls, and `3m` add/subs.

Additionally, at the end of the summations, to `acc[0]` we have added $\sum \gamma_i (1 - z_i)$ and to `acc[1]` we have added $\sum \gamma_i z_i$. But, thanks to linearity, C\sum \gamma_i (1 - z_i) = \sum \gamma_i - \sum \gamma_i z_i$.

[...]

PR #308, opened 2025-10-13
chore: add `eval_eq_batch` via Plonky3

tcoratger:
Related https://github.com/Plonky3/Plonky3/pull/1051

Should close https://github.com/tcoratger/whir-p3/issues/298

### DIFF
--- a/src/poly/evals.rs
+++ b/src/poly/evals.rs
@@ -1322,12 +1322,12 @@ mod tests {
         // Manually compute the `eq` evaluations for both points.
         let mut eq_evals_sum = vec![F::ZERO; 1 << n];
         for (p, val) in [(p1, value1), (p2, value2)] {
-            for i in 0..(1 << n) {
+            for (i, eq_eval_sum) in eq_evals_sum.iter_mut().enumerate().take(1 << n) {
                 let b0 = (i >> 1) & 1;
                 let b1 = (i >> 0) & 1;
                 let term0 = if b0 == 1 { p[0] } else { F::ONE - p[0] };
                 let term1 = if b1 == 1 { p[1] } else { F::ONE - p[1] };
-                eq_evals_sum[i] += val * term0 * term1;
+                *eq_eval_sum += val * term0 * term1;
             }
         }
 


src/whir/constraints/statement.rs
```diff
@@ -152,72 +154,45 @@ impl<F: Field> Statement<F> {
...
+        //
+        // Each thread writes to its own contiguous row, which is cache-friendly.
+        points_matrix
+            .par_rows_mut()
+            .enumerate()
+            .for_each(|(var_idx, row_slice)| {
+                for (col_idx, point) in self.points.iter().enumerate() {
+                    row_slice[col_idx] = point[var_idx];
+                }
+            });
```
SyxtonPrime: Long term, it would be great if `self.points` was also a matrix. Then you could just use the matrix transpose here,

Also, I'm not sure if there is really anything to be gained by using parallelism here. `point_matrix` is pretty small in the scheme of things, (Like 24 x 200 at the very worst) as the length of points is the number of variables not `1 << num_variables).

src/poly/evals.rs
```diff
@@ -72,7 +76,11 @@ where
     #[inline]
     pub fn accumulate(&mut self, point: &MultilinearPoint<F>, value: F) {
         assert_eq!(self.num_variables(), point.num_variables());
-        eval_eq::<_, _, true>(point.as_slice(), &mut self.0, value);
+        eval_eq_batch::<_, _, true>(
+            RowMajorMatrixView::new_col(point.as_slice()),
+            &mut self.0,
+            &[value],
+        );
```
SyxtonPrime: I think this is the wrong approch.

We should switch accumulate with a batch accumulate and then  find everywhere which uses accumulate and update to the batch version.

Currently you use accumulate inside a loop in add_new_equality

src/poly/evals.rs
```diff
@@ -85,7 +93,11 @@ where
         F: ExtensionField<BF>,
     {
         assert_eq!(self.num_variables(), point.num_variables());
-        eval_eq_base::<_, _, true>(point.as_slice(), &mut self.0, value);
+        eval_eq_base_batch::<_, _, true>(
+            RowMajorMatrixView::new_col(point.as_slice()),
+            &mut self.0,
+            &[value],
+        );
```
SyxtonPrime: Similarly this function should also be updated to a batch version.

SyxtonPrime (approved):
Mostly looks good. A couple of minor comments.

src/whir/constraints/statement.rs
tcoratger: True:
- I'll open an issue to transform `self.points` into a matrix.
- I've tested and indeed the parallelism doesn't bring much here so I've just removed it.

src/poly/evals.rs
tcoratger: Fixed

PR #315, opened 2025-10-27
dft: rm from public api

tcoratger:
Should supersede https://github.com/tcoratger/whir-p3/pull/314

@kilic what do you think about this?

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -118,13 +118,14 @@ fn main() {
         folding_factor,
         merkle_hash,
         merkle_compress,
+        dft: Radix2DFTSmallBatch::<F>::default(),
         soundness_type,
         starting_log_inv_rate: starting_rate,
         rs_domain_initial_reduction_factor,
         univariate_skip: false,
     };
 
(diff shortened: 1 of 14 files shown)


src/bin/main.rs
```diff
@@ -118,13 +118,14 @@ fn main() {
         folding_factor,
         merkle_hash,
         merkle_compress,
+        dft: Radix2DFTSmallBatch::<F>::default(),
```
kilic: It becomes a bit faster on my end when we configure the fft instance like
```rust
 let max_fft_size = num_variables + starting_rate - folding_factor.at_round(0);
 let dft = Radix2DFTSmallBatch::<F>::new(1 << max_fft_size);
 ```

tcoratger: Yes you are right (I've the same locally), let us not merge this for now since we don't want to degrade performance for now, especially for such a detail.

I've another idea that would unblock things and would be to add a new method `new(size:usize)` inside this trait https://github.com/Plonky3/Plonky3/blob/658115aae28312e97cd3ef38d1ab046839f734a8/dft/src/traits.rs#L27. This would allow us to just rely on this and be able to init any DFT. Will check this.

PR #321, opened 2025-10-30
Reverse variable order

kilic:
In this PR variable order of sumcheck randomness is reversed. So that we can fold running polynomial without requiring extra memory. As a result I'm observing nice speed up in sumcheck rounds. Also now sumcheck operations should become more SIMD friendly.

However I had to add transpose step before the encoding to get reversed codeword and it takes %15 of commitment time.

Some tests are not passing now since 'univariate skip' parts are not integrated yet, so I leave this PR as draft for now.

### DIFF
--- a/src/whir/prover/mod.rs
+++ b/src/whir/prover/mod.rs
@@ -397,27 +397,6 @@ where
             Some(constraint),
         );
 
-        let start_idx = self.folding_factor.total_number(round_index);
-        let dst_randomness =
-            &mut round_state.randomness_vec[start_idx..][..folding_randomness.num_variables()];
-
-        // Store randomness: for skip case, don't reverse to maintain correct order
-        if round_state.used_univariate_skip {
-            for (dst, src) in dst_randomness
-                .iter_mut()
-                .zip(folding_randomness.iter().copied())
-            {
-                *dst = src;
-            }
-        } else {
-            for (dst, src) in dst_randomness
-                .iter_mut()
-                .zip(folding_randomness.iter().rev())
-            {
-                *dst = *src;
-            }
-        }
-
         // Update round state
         round_state.domain_size = new_domain_size;
         round_state.next_domain_gen =
@@ -530,33 +509,12 @@ where
 
         // Run final sumcheck if required
         if self.final_sumcheck_rounds > 0 {
-            let final_folding_randomness =
-                round_state.sumcheck_prover.compute_sumcheck_polynomials(
-                    prover_state,
-                    self.final_sumcheck_rounds,
-                    self.final_folding_pow_bits,
-                    None,
-                );
-            let start_idx = self.folding_factor.total_number(round_index);
-            let rand_dst = &mut round_state.randomness_vec
-                [start_idx..start_idx + final_folding_randomness.num_variables()];
-
-            // Store final randomness: for skip case, don't reverse to maintain correct order
-            if round_state.used_univariate_skip {
-                for (dst, src) in rand_dst
-                    .iter_mut()
-                    .zip(final_folding_randomness.iter().copied())
-                {
-                    *dst = src;
-                }
-            } else {
-                for (dst, src) in rand_dst
-                    .iter_mut()
-                    .zip(final_folding_randomness.iter().rev())
-                {
-                    *dst = *src;
-                }
-            }
+            round_state.sumcheck_prover.compute_sumcheck_polynomials(
+                prover_state,
+                self.final_sumcheck_rounds,
+                self.final_folding_pow_bits,
+                None,
+            );
         }
 
         Ok(())
--- a/src/whir/prover/round_state/state.rs
+++ b/src/whir/prover/round_state/state.rs
@@ -103,20 +103,7 @@ where
     ///
     /// This affects how randomness is stored across rounds to maintain consistency
     /// with the constraint evaluation format.
-    pub used_univariate_skip: bool,
     pub merkle_prover_data: Option<RoundMerkleTree<F, EF, W, DIGEST_ELEMS>>,
-
-    /// Global randomness vector tracking all folding challenges across rounds.
-    ///
-    /// This vector accumulates folding randomness in reverse variable order:
-    /// position i stores the challenge for variable X_{n-1-i}. As rounds progress,
-    /// new challenges are prepended, building the complete evaluation point needed
-    /// for final constraint verification.
-    ///
-    /// The vector enables efficient polynomial evaluation at the accumulated
-    /// challenge point while maintaining the proper variable ordering for
-    /// multilinear extensions and sumcheck interactions.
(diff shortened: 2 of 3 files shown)


src/poly/evals.rs
```diff
@@ -232,27 +235,17 @@ where
...
-            }
-            self.0.truncate(mid);
-        }
+        let mid = self.0.len() / 2;
+
+        // Evaluations at `a_i` and `a_{i + n/2}` slots are folded with `r` into `a_i` slot
+        let (p0, p1) = self.0.split_at_mut(mid);
+        p0.par_iter_mut()
+            .zip(p1.par_iter())
+            .with_min_len(PARALLEL_THRESHOLD)
```
tcoratger: I think `with_min_len` is available on Rayon, but since we're using Plonky3's `maybe_rayon` to adapt to the user's configuration (between a cfg that allows parallelization and one that doesn't and to easily integrate to Plonky3), this isn't available. That being said, if this is a feature we absolutely need, we can add it to Plonky3 as well.

src/poly/evals.rs
```diff
@@ -269,23 +262,19 @@ where
...
-
-        EvaluationsList::new(folded)
+        let mid = self.0.len() / 2;
+        // Evaluations at `a_i` and `a_{i + n/2}` slots are folded with `r`
+        let (p0, p1) = self.0.split_at(mid);
+        // Create new EvaluationsList in the extension field
+        EvaluationsList(
+            p0.par_iter()
+                .zip(p1.par_iter())
+                .with_min_len(PARALLEL_THRESHOLD)
```
tcoratger: Same comment about `with_min_len`

src/poly/multilinear.rs
```diff
@@ -169,6 +169,12 @@ where
         // Evaluate the new, smaller polynomial at the remaining challenges `r_rest`.
         EvaluationsList::new(folded_row).evaluate_hypercube(&r_rest)
     }
+
+    pub fn reversed(&self) -> Self {
+        let mut vars = self.0.clone();
+        vars.reverse();
+        Self(vars)
+    }
```
tcoratger: Does this work?

```suggestion
    #[must_use]
    pub fn reversed(&self) -> Self {
        let vars = self.0.iter().rev().copied().collect();
        Self(vars)
    }
```

src/sumcheck/tests.rs
```diff
@@ -233,18 +233,11 @@ where
...
-        .chain(point.iter())
-        .copied()
-        .collect();
-    MultilinearPoint::new(new_coords)
+impl<F: Field> MultilinearPoint<F> {
+    /// Helper to extend a `MultilinearPoint` by creating a new one.
+    pub fn extend(&mut self, other: &MultilinearPoint<F>) {
+        self.0.extend_from_slice(&other.0);
+    }
 }
```
tcoratger: Maybe we can just implement this in the `MultilinearPoint` file under a test cfg. You never know, it might be useful here; it could also be useful for other purposes.

tcoratger (commented):
That's much better, even in terms of clarity. It removes the reverse operations scattered everywhere, which were complicated to understand, and we were getting a bit confused between the reverse operations going in both directions.

Now I guess the next steps are:

- To solve the minor compilation issues.

- To make the `test_whir_end_to_end_without_univariate_skip` and `test_whir_with_initial_statement_false` tests pass. This will mean that the basic setup is working and that most of the work is already done.

- Then we'll need to look at `test_whir_end_to_end_with_univariate_skip`, probably with a minimal parameter configuration. For example, I usually use this to debug univariate skip:

```rust
let num_variable = 6;
let folding_factor = FoldingFactor::ConstantFromSecondRound(5, 2);
let num_points = 0;
let soundness_type = SecurityAssumption::JohnsonBound;
let pow_bits = 0;
let rs_domain_initial_reduction_factor = 1;

make_whir_things(
    num_variable,
    folding_factor,
    num_points,
    soundness_type,
    pow_bits,
    rs_domain_initial_reduction_factor,
    true,
    true,
);
```

In this case, the problem will most likely come from a `if is_univariate_skip` block with a missing reverse statement. It is also possible to check in this file https://github.com/tcoratger/whir-p3/blob/main/src/sumcheck/sumcheck_single_skip.rs that the change of order does not cause any problems, in particular with the help of this test https://github.com/tcoratger/whir-p3/blob/f91423911e8f51484342857f1719220e37f59e0c/src/sumcheck/sumcheck_single_skip.rs#L214-L565 which is very manual and which helps to verify that we are doing exactly what is expected for a minimal case. This will help, for example, to visualize the inputs and the function path to ensure that there is no problem with this change of order of variables which would lead, for example, to a transpose or a change of order of lines or any other operation which should be changed inside `compute_skipping_sumcheck_polynomial`.

src/whir/constraints/evaluator.rs
```diff
@@ -161,11 +155,10 @@ impl ConstraintPolyEvaluator {
                     // evaluation, or a standard evaluation.
                     let val = if is_skip_round {
                         // Skip-aware evaluation over r_rest || r_skip.
-                        debug_assert_eq!(point.num_variables(), self.num_variables);
+                        // TODO use self.univariate_skip.unwrap()
                         point.eq_poly_with_skip(&point_for_round, K_SKIP_SUMCHECK)
```
tcoratger: Here this is possible there is something to check related to `point_for_round` ordering because here


https://github.com/tcoratger/whir-p3/blob/d3d2d2e66ba95e38f2cc398b41b5d0a04c0a0819/src/poly/multilinear.rs#L285-L308

we are building a new `EvaluationsList` and so I think that to do that, `point_for_round` should be in the proper lex order so I wouldn't be surprised if `point_for_round` had to be reversed before entering the function (again not sure, just a possibility).

src/sumcheck/sumcheck_single.rs
```diff
@@ -146,15 +144,18 @@ pub(crate) fn compute_sumcheck_polynomial<F: Field, EF: ExtensionField<F>>(
...
+    let mid = evals.num_evals() / 2;
+    let (plo, phi) = evals.0.split_at(mid);
+    let (elo, ehi) = weights.0.split_at(mid);
+
+    let (c0, c2) = plo
+        .par_iter()
+        .zip(phi.par_iter())
+        .zip(elo.par_iter().zip(ehi.par_iter()))
+        .map(|((&p0, &p1), (&e0, &e1))| (e0 * p0, (e1 - e0) * (p1 - p0)))
+        .reduce(
```
tcoratger: Here I think that you have to use `par_fold_reduce` to be fully compatible like here

https://github.com/tcoratger/whir-p3/blob/d3d2d2e66ba95e38f2cc398b41b5d0a04c0a0819/src/sumcheck/sumcheck_single.rs#L149-L158

PR #325, opened 2025-11-06
feat: add no_std support

### LINKED ISSUE #324: chore: make the code no_std compatible
In order to upstream the code easily in Plonky3 in the future, we need it to be no_std compatible. 

As a consequence we should have `#![no_std]` in our lib.rs file and propagate the changes all over the codebase so that there is no std dependency anymore.

Nashtare:
closes #324 

Kept the requirement as a conditional attribute because of the timing tests `utils` module (but could be tweaked / removed if we wanted a global `#![no_std]` including tests).

Relies on `alloc` and `hashbrown` for collections.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: nightly
-          targets: ${{ env.target }}
+          target: ${{ env.target }}
       
       - name: Build no-std
         run: cargo build --verbose --target ${{ env.target }} --no-default-features


Nashtare:
@tcoratger done :)

tcoratger (approved):
Looks good to me, just the embedded compilation thing failing in the CI (I guess because the target isn’t installed for the toolchain our CI is using, not sure).

Then we are good to merge.

I'll open another followup PR to fix the thing with the test randomness, like this is done in Plonky3.

.github/workflows/rust.yml
```diff
@@ -28,6 +28,23 @@ jobs:
...
+      
+    env:
+      target: thumbv7em-none-eabi
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install nightly toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: nightly
```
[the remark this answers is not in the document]
Nashtare: I've used `targets: ${{ env.target }}` for consistency and in case this was to be changed later on, but should be good now (I hope!).

.github/workflows/rust.yml
```diff
@@ -28,6 +28,24 @@ jobs:
...
+          targets: ${{ env.target }}
```
tcoratger: Maybe this (singular not `targets`)? At least ChatGPT told me this... CI jobs are a pain...

```suggestion
          target: ${{ env.target }}
```

Nashtare: hmm I believe it should be `targets`. Let me try a different approach

Nashtare: Sure can do

PR #348, opened 2025-11-14
Whir Proof Structure

camofu:
Introduces proof.rs containing the complete data structure definitions for WHIR proofs


tcoratger:
@CarloModicaPortfolio Sorry, I had to revert it here: https://github.com/tcoratger/whir-p3/pull/349.

In fact, you didn't add `pub mod proof;` in the `mod.rs` file, so it wasn't compiling the `proof.rs` file. Adding it locally, I realized there were quite a few compilation errors, so this pull request needs to be redone by including this and fixing the compilation errors. It should be quite easy.

Issue #279, opened 2025-09-09
Compare WHIR and FRI


tcoratger:
### Description

Currently, in the [main file](https://github.com/tcoratger/whir-p3/blob/main/src/bin/main.rs), WHIR can be run fully with:

```bash
RUSTFLAGS='-C target-cpu=native' cargo run --release
````

This includes:

* Commit phase https://github.com/tcoratger/whir-p3/blob/14fdffe2ad6d1018d73e8426eba25a86e2513e56/src/bin/main.rs#L171-L173
* Prover phase https://github.com/tcoratger/whir-p3/blob/14fdffe2ad6d1018d73e8426eba25a86e2513e56/src/bin/main.rs#L181-L183
* Verify phase https://github.com/tcoratger/whir-p3/blob/14fdffe2ad6d1018d73e8426eba25a86e2513e56/src/bin/main.rs#L202-L204

### Feature request

It would be useful to add a **Plonky3 FRI run** to the same file, so that we have a direct, apples-to-apples comparison between WHIR and FRI.
The configuration should remain identical (same polynomial, same hashes, etc.) to ensure a fair benchmark.

For reference, the Plonky3 repo has an example here: https://github.com/Plonky3/Plonky3/blob/main/examples/src/proofs.rs that could serve as inspiration.

### Motivation

Having WHIR and FRI run side-by-side in the same environment will allow us to compare:

* Prover time
* Verifier time
* Proof size

This comparison will help us track where WHIR stands relative to FRI under realistic conditions.

PR #366, opened 2025-11-21
Add FRI comparison, closes issue #279

Vindaar:
This PR adds a comparison to the FRI PCS in the main binary. Due to the very different nature of FRI vs WHIR (multilinear vs univariate), I'm not certain how apples-to-apples this really is.

In addition I still have a few open questions:
- pretty uncertain if I should use `TwoAdicMultiplicativeCoset` or some other domain?
- do I handle the `RowMajorMatrix` construction sensibly? I split the evaluations sampled for WHIR into `NUM_COLS`. Currently a `const`, but we could make that a command line argument.
- The two adicity of KoalaBear is 24. This means the longest "trace height" supported is 2^23. The default parameters sets `num_variables` to 25. At `NUM_COLS = 1`, this would fail.
- I'm uncertain about how I construct the verifier points to be honest.
- currently we don't use a packing variant of Poseidon2. Should we? 

With the parameters I currently hardcoded, I get this output:

```
Proving time: 1694 ms (commit: 786 ms, opening: 908 ms)
proof size: 106.95 KiB
Verification time: 2636 μs

=========================================
FRI (PCS) 🍳️
.........
Proof Verified Successfully

Proving time: 1059 ms (commit: 641 ms, opening: 418 ms)
proof size: 315.26 KiB
Verification time: 8199 μs
```

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -391,11 +391,8 @@ fn run_fri(ctx: &mut Context) {
     let points = vec![open_points.clone(); num_chunks];
     // Generate the opening proof
     let open_time = Instant::now();
-    let (opened_values, proof) = Pcs::<EF, MyChallenger>::open(
-        &pcs,
-        vec![(&prover_data, points)],
-        &mut ctx.challenger.clone(),
-    );
+    let (opened_values, proof) =
+        pcs.open(vec![(&prover_data, points)], &mut ctx.challenger.clone());
     let opening_time = open_time.elapsed();
 
     // Construct the points needed for the verifier
@@ -413,8 +410,7 @@ fn run_fri(ctx: &mut Context) {
 
     // Verify the opening proof
     let verif_time = Instant::now();
-    let res = Pcs::<EF, MyChallenger>::verify(
-        &pcs,
+    let res = pcs.verify(
         vec![(commitment, verifier_points)],
         &proof,
         &mut ctx.challenger.clone(),


WizardOfMenlo:
Not sure how hard it is to do, but if you want to compare FRI and WHIR apple to apple (as low-degree tests) you can remove the following from WHIR (either properly or just from the timings)
- The OOD sample in the commit phase.
- The initial sumcheck in WHIR (basically, you only require a sumcheck after the first prover oracle is sent).

(You can look at https://github.com/WizardOfMenlo/whir/blob/15cf6668e904ed2e80c9e6209dcce69f5bcf79b9/src/bin/main.rs#L195 for an example, I think it's the `initial_statement: false` option therein).

tcoratger:
@WizardOfMenlo @Vindaar 

- Yes I think that right now WHIR is running as a Polynomial Commitment Scheme while FRI is running as a Low-Degree Test.

  A Low-Degree Test verifies that a function $f: L \to \mathbb{F}$ is close to a low-degree polynomial. So in this case, inputs are evaluations of $f$ over domain $L$ and the verifier accept if $f$ is $\delta$-close to a degree-$d$ polynomial, reject otherwise. But there is no evaluation constraints, it only tests proximity to the code.
  
  On the other hand, a PCS allows a prover to commit to a polynomial and later open it at specific points. So, the prover commits to polynomial $f$ and later prover reveals $f(z) = y$ for verify-chosen point $z$. So it uses LDT as a building block but adds evaluation checking.
  
  The fact that you have `initial_statement: true` is what makes WHIR run as PCS, not LDT. It includes an initial sumcheck to verify evaluations and adds out-of-domain (OOD) checks during commit.
  
  But indeed, by marking this flag as false (doesn't check the evaluation constraints), you will avoid the first sumcheck and thus save significant costs.
  
  I believe you can take inspiration from the link sent by Giacomo on the other WHIR repo, but I think it's desirable to keep a main run with WHIR as PCS in parallel (in addition to the comparison with FRI) because that's what interests us most. Therefore, to get an idea of ​​the performance improvements, it's always useful to be able to run it as PCS quickly with the CLI.

- It also seems that you have a much higher security level for WHIR. The number of queries for FRI (40) and the 8 bits of grinding seems much lower than the 90 bits of security imposed by default in the WHIR cli no? So in my sense, with the current configuration, FRI requires fewer queries, smaller proofs and faster verification due to this mismatch.

  We should probably have something like this to match (not sure about my calculations, we need to make this clean and document where the calculations come from so that the reader is sure to understand):
  
  ```rust
    let target_security = 90;
    let fri_queries = target_security - pow_bits;
    
    FriParameters {
        log_blowup: 1,
        num_queries: fri_queries, 
        proof_of_work_bits: pow_bits,
        // ...
    }
   ```

- The matrix reshaping thing looks a bit strange to me. If I understand correctly:
  1. We take a univariate polynomial with `2^25` evaluations
  2. We arbitrarily split it into 32 columns
  3. Reduces domain size from `2^25` to `2^20`
  4. Creates a `2^20 × 32` matrix
  
  Why 32 columns? It seems that the domain size `2^20` instead of `2^25` changes the problem, so it's not exactly clear to me why we're doing this. I'd like to be sure that by reshaping the problem in this way, we're not changing the conditions and therefore comparing apples to apples (maybe if we just do `const NUM_COLS: usize = 1;` that's enough, right?).

- I also think we're using evaluation point types that aren't comparable with

  ```rust
  let points: Vec<_> = (0..ctx.num_evaluations)
      .map(|_| MultilinearPoint::rand(&mut ctx.rng, params.num_variables))
      .collect();
  
  let open_points: Vec<EF> = (0..ctx.num_evaluations)
      .map(|_| ctx.rng.random::<EF>())
      .collect();
  ```
  
  But migrating to the LDT comparison should remove this because in this case, we're not doing evaluation checks at all; we're just testing proximity.

- I also think the comparison of proof sizes isn't exactly fair.

  - WHIR: Calculates from field elements: `num_elements * bits_per_element / 8`
  - FRI: Uses `bincode` serialization which adds overhead (length prefixes, metadata)


- It's minor but I think that in the actual comparison, it is not exactly clear what the relationship is between WHIR's multilinear degree and FRI's univariate degree is not explicit.

  WHIR:
  - `num_variables = 25`
  - Multilinear polynomial: $\hat{f}(x_1, \ldots, x_{25})$
  - Degree: 1 in each variable
  - Total evaluations: $2^{25}$
  - Can be viewed as a univariate of degree $2^{25} - 1$
  
  FRI (current setup):
  - Domain: $2^{20}$ points with 32 columns
  - Unclear what the actual univariate degree being tested is

  We should document this more clearly in the code so the reader can understand exactly how we are trying to compare univariate and multilinear functions.

src/bin/main.rs
```diff
@@ -231,3 +295,152 @@ fn main() {
...
+        commitment_time.as_millis() + opening_time.as_millis(),
+        commitment_time.as_millis(),
+        opening_time.as_millis()
+    );
+    let proof_size = calc_fri_proof_size(opened_values, proof) as f64;
+    println!("proof size: {:.2} KiB", proof_size / 1024.0);
+    println!("Verification time: {} μs", verify_time.as_micros());
+}
+
+#[allow(clippy::too_many_lines)]
```
tcoratger:
```suggestion
```

src/bin/main.rs
```diff
@@ -231,3 +295,152 @@ fn main() {
...
+/// These parameters represents numbers used in Valida
+pub const fn create_benchmark_fri_params<Mmcs>(mmcs: Mmcs) -> FriParameters<Mmcs> {
+    FriParameters {
+        log_blowup: 1,
+        log_final_poly_len: 0,
+        num_queries: 40,
+        proof_of_work_bits: 8,
+        mmcs,
+    }
+}
```
SyxtonPrime: Ideally, we should fix a security parameter and then calculate the number of queries based off that.

src/bin/main.rs
```diff
@@ -231,3 +295,152 @@ fn main() {
...
+        .collect();
+    let num_chunks = ctx.polynomial.num_evals() / trace_height;
+    let points = vec![open_points.clone(); num_chunks];
+    // Generate the opening proof
+    let open_time = Instant::now();
+    let (opened_values, proof) = Pcs::<EF, MyChallenger>::open(
+        &pcs,
+        vec![(&prover_data, points)],
+        &mut ctx.challenger.clone(),
+    );
```
SyxtonPrime:
```suggestion
    let (opened_values, proof) = pcs.open(
        vec![(&prover_data, points)],
        &mut ctx.challenger.clone(),
    );
```

src/bin/main.rs
```diff
@@ -231,3 +295,152 @@ fn main() {
...
+        .collect();
+
+    // Verify the opening proof
+    let verif_time = Instant::now();
+    let res = Pcs::<EF, MyChallenger>::verify(
+        &pcs,
+        vec![(commitment, verifier_points)],
+        &proof,
+        &mut ctx.challenger.clone(),
+    );
```
SyxtonPrime:
```suggestion
    let res = pcs.verify(
        vec![(commitment, verifier_points)],
        &proof,
        &mut ctx.challenger.clone(),
    );
```

SyxtonPrime:
Also I'm pretty sure you are using the packing version of Poseidon2. This should be used automatically due to how you've defined `Poseidon2MerkleMmcs`.

src/bin/main.rs
Vindaar: Huh, this does not actually produce the same problems as for `commit`. I thought I saw similar errors, but I guess that was while I was still writing code or I misremember.

src/bin/main.rs
Vindaar: Also works to my surprise.

src/bin/main.rs
MuhtasimTanmoy:
```rust
let mut proof = WhirProof::<F, EF, 8>::default();
```

Should be

```rust
let mut proof = WhirProof::<F, EF, F, 8>::from_protocol_parameters(&params, params.num_variables);
```

PR #371, opened 2025-11-26
round state: use proof initial phase to dispatch in sumcheck

tcoratger:
Should supersede https://github.com/tcoratger/whir-p3/pull/370

### DIFF
--- a/src/whir/prover/round_state/tests.rs
+++ b/src/whir/prover/round_state/tests.rs
@@ -165,15 +165,11 @@ fn test_no_initial_statement_no_sumcheck() {
     // - domain separator for Fiat-Shamir transcript,
     // - prover state,
     // - witness containing Merkle tree for `poly`.
-    let (_, _, mut prover_state, witness) = setup_domain_and_commitment(&config, poly);
+    let (mut proof, mut challenger, mut prover_state, witness) =
+        setup_domain_and_commitment(&config, poly);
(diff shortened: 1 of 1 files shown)


src/sumcheck/sumcheck_single.rs
```diff
@@ -384,10 +380,12 @@ where
...
     #[instrument(skip_all)]
     #[allow(clippy::too_many_arguments)]
-    pub fn with_skip<Challenger, const DIGEST_ELEMS: usize>(
+    pub fn with_skip<Challenger>(
         evals: &EvaluationsList<F>,
         prover_state: &mut ProverState<F, EF, Challenger>,
-        proof: &mut WhirProof<F, EF, DIGEST_ELEMS>,
+        skip_evaluations: &mut Vec<EF>,
+        skip_pow: &mut Option<F>,
+        sumcheck: &mut SumcheckData<EF, F>,
```
CarloModicaPortfolio: we should define in the future a structu with 
define a struct that has inside 

```rust
        skip_evaluations: &mut Vec<EF>,
        skip_pow: &mut Option<F>,
        sumcheck: &mut SumcheckData<EF, F>,
```
when we are going to remove the` #[allow(clippy::too_many_arguments)]`

CarloModicaPortfolio (approved):

PR #378, opened 2025-12-02
Guard prox_gaps_error against invalid num_functions

adust09:
Previously `prox_gaps_error` accepted any `num_functions` and would compute `log2(num_functions - 1)`, yielding NaN/∞ if 1 or 0 slipped in. 

This PR adds a guard asserting `num_functions >= 2`, matching the intended multi-function domain and the current callers (all pass 2), so misuse fails fast instead of corrupting soundness estimates.

### DIFF
--- a/src/parameters/errors.rs
+++ b/src/parameters/errors.rs
@@ -250,6 +250,20 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


tcoratger (approved):
Thanks a lot :)

PR #383, opened 2025-12-03
SIMD optimisations

kilic:
This PR adds SIMD optimisations for sumcheck and multilinar evaluation. I observe 15-20% performance improvement in opening time in `main` demo.

Notes: 

* To get packed combination output I replaced `p3_multilinar_utils::batch_eq` with a new implementation that performs similar with the previous one.
* To encode and commit the running sumcheck polynomial, we unpack the packed polynomial after running folding factor number of rounds. This adds only a marginal cost.
* We may consider removing the generic eval_multilinear function and instead favor specialized implementations for the base and extension fields. I kept it for now to demonstrate the benchmarks.

### DIFF
--- a/src/poly/evals.rs
+++ b/src/poly/evals.rs
@@ -35,11 +35,11 @@ pub struct EvaluationsList<F>(pub(crate) Vec<F>);
 impl<F: Copy + Clone + Send + Sync> EvaluationsList<F> {
     /// Given a number of points initializes a new zero polynomial
     #[inline]
-    pub fn zero(num_evals: usize) -> Self
+    pub fn zero(num_variables: usize) -> Self
     where
         F: PrimeCharacteristicRing,
     {
-        Self(F::zero_vec(1 << num_evals))
+        Self(F::zero_vec(1 << num_variables))
     }
 
     /// Constructs an `EvaluationsList` from a vector of evaluations.
@@ -405,9 +405,10 @@ where
         point: &MultilinearPoint<EF>,
     ) -> EF {
         if point.num_variables() < MLE_RECURSION_THRESHOLD {
-            return eval_multilinear_recursive(&self.0, point.as_slice());
+            eval_multilinear_recursive(&self.0, point.as_slice())
+        } else {
+            eval_multilinear_base::<F, EF>(&self.0, point.as_slice())
         }
-        eval_multilinear_base::<F, EF>(&self.0, point.as_slice())
     }
 
     /// Evaluates the multilinear polynomial at `point ∈ F^n`.
@@ -427,9 +428,10 @@ where
         F: ExtensionField<BaseField>,
     {
         if point.num_variables() < MLE_RECURSION_THRESHOLD {
-            return eval_multilinear_recursive(&self.0, point.as_slice());
+            eval_multilinear_recursive(&self.0, point.as_slice())
+        } else {
+            eval_multilinear_ext::<BaseField, F>(&self.0, point.as_slice())
         }
-        eval_multilinear_ext::<BaseField, F>(&self.0, point.as_slice())
     }
 
     /// Folds a multilinear polynomial stored in evaluation form along the last `k` variables.
@@ -998,7 +1000,9 @@ impl<F> IntoIterator for EvaluationsList<F> {
     }
 }
 
-fn eval_multilinear_small<F, EF>(evals: &[F], point: &[EF]) -> EF
+/// Evaluates a multilinear polynomial `evals` at `point` using a recursive strategy.
+/// For small numbers of variables (<=4) it switches to the unrolled strategy.
+fn eval_multilinear_recursive<F, EF>(evals: &[F], point: &[EF]) -> EF
 where
     F: Field,
     EF: ExtensionField<F>,
@@ -1007,8 +1011,6 @@ where
     //
     // This is a critical invariant: `evals.len()` must be exactly `2^point.len()`.
     debug_assert_eq!(evals.len(), 1 << point.len());
-    // This method is only for small number of variables (<=4)
-    debug_assert!(point.len() <= 4);
 
     // Select the optimal evaluation strategy based on the number of variables.
     match point {
@@ -1065,58 +1067,39 @@ where
             let a1 = a10 + *x1 * (a11 - a10);
             a0 + (a1 - a0) * *x0
         }
-        _ => {
-            unreachable!()
+        // General Case (5+ Variables)
+        //
+        // This handles all other cases, using one of two different strategies.
+        [x, sub_point @ ..] => {
+            // Split the evaluations into two halves, corresponding to the first variable being 0 or 1.
+            let (f0, f1) = evals.split_at(evals.len() / 2);
+
+            // Recursively evaluate on the two smaller hypercubes.
+            let (f0_eval, f1_eval) = {
+                // Only spawn parallel tasks if the subproblem is large enough to overcome
+                // the overhead of threading.
+                let work_size: usize = (1 << 15) / core::mem::size_of::<F>();
+                if evals.len() > work_size {
+                    join(
+                        || eval_multilinear_recursive(f0, sub_point),
+                        || eval_multilinear_recursive(f1, sub_point),
+                    )
+                } else {
+                    // For smaller subproblems, execute sequentially.
+                    (
+                        eval_multilinear_recursive(f0, sub_point),
+                        eval_multilinear_recursive(f1, sub_point),
+                    )
+                }
+            };
+            // Perform the final linear interpolation for the first variable `x`.
+            f0_eval + (f1_eval - f0_eval) * *x
         }
     }
 }
 
-pub fn eval_multilinear_recursive<F, EF>(evals: &[F], point: &[EF]) -> EF
-where
-    F: Field,
-    EF: ExtensionField<F>,
-{
-    // Ensure that the number of evaluations matches the number of variables in the point.
-    //
-    // This is a critical invariant: `evals.len()` must be exactly `2^point.len()`.
-    let num_vars = point.len();
-    debug_assert_eq!(evals.len(), 1 << num_vars);
-
-    // Select the optimal evaluation strategy based on the number of variables.
-    if num_vars <= 4 {
-        return eval_multilinear_small(evals, point);
-    }
-    // Get the current variable
-    let x = &point[0];
-    // Create a new point with the remaining coordinates.
-    let sub_point = &point[1..num_vars];
-
-    // Split the evaluations into two halves, corresponding to the first variable being 0 or 1.
-    let (f0, f1) = evals.split_at(evals.len() / 2);
-
-    // Recursively evaluate on the two smaller hypercubes.
-    let (f0_eval, f1_eval) = {
-        // Only spawn parallel tasks if the subproblem is large enough to overcome
-        // the overhead of threading.
-        let work_size: usize = (1 << 15) / core::mem::size_of::<F>();
-        if evals.len() > work_size {
-            join(
-                || eval_multilinear_recursive(f0, sub_point),
-                || eval_multilinear_recursive(f1, sub_point),
-            )
-        } else {
-            // For smaller subproblems, execute sequentially.
-            (
-                eval_multilinear_recursive(f0, sub_point),
-                eval_multilinear_recursive(f1, sub_point),
-            )
-        }
-    };
-    // Perform the final linear interpolation for the first variable `x`.
-    f0_eval + (f1_eval - f0_eval) * *x
-}
-
-pub fn eval_multilinear_base<F, EF>(evals: &[F], point: &[EF]) -> EF
+/// Evaluates a multilinear polynomial `evals` at `point` where `evals` are in the base field `F` and `point` is in the extension field `EF`.
+fn eval_multilinear_base<F, EF>(evals: &[F], point: &[EF]) -> EF
 where
     F: Field,
     EF: ExtensionField<F>,
@@ -1163,7 +1146,8 @@ where
     EF::ExtensionPacking::to_ext_iter([sum]).sum()
 }
 
-pub fn eval_multilinear_ext<F, EF>(evals: &[EF], point: &[EF]) -> EF
+/// Evaluates a multilinear polynomial `evals` at `point` where `evals` and `point` are in the extension field `EF`.
+fn eval_multilinear_ext<F, EF>(evals: &[EF], point: &[EF]) -> EF
 where
     F: Field,
     EF: ExtensionField<F>,
@@ -1227,7 +1211,7 @@ mod tests {
     type F = BabyBear;
     type EF4 = BinomialExtensionField<F, 4>;
 
-    // Naive method to evaluate multilinear polynomials for testing.
+    /// Naive method to evaluate a multilinear polynomial for testing.
     fn eval_multilinear<F: Field, EF: ExtensionField<F>>(evals: &[F], point: &[EF]) -> EF {
         let eq = EvaluationsList::new_from_point(point, EF::ONE);
         dot_product(eq.iter().copied(), evals.iter().copied())
@@ -3413,18 +3397,14 @@ mod tests {
         let mut rng = SmallRng::seed_from_u64(1);
 
         for k in 0..=18 {
-            let list: Vec<F> = (0..(1 << k)).map(|_| rng.random()).collect();
+            let poly: Vec<F> = (0..1 << k).map(|_| rng.random()).collect();
             let point = MultilinearPoint::<EF4>::rand(&mut rng, k);
             let point = point.as_slice();
-            let e0 = eval_multilinear_recursive(&list, point);
-            let e1 = eval_multilinear(&list, point);
+            let e0 = eval_multilinear_recursive(&poly, point);
+            let e1 = eval_multilinear(&poly, point);
             assert_eq!(e0, e1);
-            let e1 = eval_multilinear_base(&list, point);
+            let e1 = eval_multilinear_base(&poly, point);
             assert_eq!(e0, e1);
-            if list.len() < 4 {
-                let e1 = eval_multilinear_small(&list, point);
-                assert_eq!(e0, e1);
-            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/poly/evals.rs
```diff
@@ -52,11 +63,158 @@ where
...
-    /// Given a number of points initializes a new zero polynomial
-    pub fn zero(num_evals: usize) -> Self {
-        Self(F::zero_vec(1 << num_evals))
+    /// Evaluates the polynomial as a constant.
+    /// This is only valid for constant polynomials (i.e., when `num_variables` is 0).
+    ///
+    /// Returns None in other cases.
+    ///
+    /// # Panics
+    /// Panics if `num_variables` is not 0.
```
tcoratger: I think there is no panic here (that was there before so this is not linked to this PR)

```suggestion
```

src/poly/evals.rs
```diff
@@ -52,11 +63,158 @@ where
...
+    ///
+    /// The compression is achieved by applying the following formula to pairs of evaluations:
+    /// ```text
+    ///     p'(X_2, ..., X_n) = (p(1, X_2, ..., X_n) - p(0, X_2, ..., X_n)) \cdot r + p(0, X_2, ..., X_n)
+    /// ```
+    pub fn compress_packed<Ext>(&self, zi: Ext) -> EvaluationsList<Ext::ExtensionPacking>
+    where
+        A: Field,
+        Ext: ExtensionField<A>,
+    {
```
tcoratger: To match the header documentation

```suggestion
    pub fn compress_packed<EF>(&self, r: EF) -> EvaluationsList<EF::ExtensionPacking>
    where
        A: Field,
        EF: ExtensionField<A>,
    {
```

src/poly/evals.rs
```diff
@@ -52,11 +63,156 @@ where
...
+        self.0.truncate(mid);
+    }
+
+    /// Folds a list of evaluations from a base field `F` into packed form of extension field `EF`.
+    ///
+    /// ## Arguments
+    /// * `r`: A value `r` from the extension field `EF`, used as the random challenge for folding.
+    ///
+    /// ## Returns
+    /// A new `EvaluationsList<EF>` containing the compressed evaluations in the extension field.
```
tcoratger:
```suggestion
    /// ## Returns
    /// A new `EvaluationsList<Ext::ExtensionPacking>` containing the compressed evaluations in
    /// packed form.
```

src/poly/evals.rs
```diff
@@ -52,11 +63,156 @@ where
...
+    /// The compression is achieved by applying the following formula to pairs of evaluations:
+    /// ```text
+    ///     p'(X_2, ..., X_n) = (p(1, X_2, ..., X_n) - p(0, X_2, ..., X_n)) \cdot r + p(0, X_2, ..., X_n)
+    /// ```
+    pub fn compress_packed<Ext>(&self, zi: Ext) -> EvaluationsList<Ext::ExtensionPacking>
+    where
+        A: Field,
+        Ext: ExtensionField<A>,
+    {
+        let zi = Ext::ExtensionPacking::from_ext_slice(&vec![zi; A::Packing::WIDTH]);
```
tcoratger: Not sure this works but something like this or with `from_fn` would be nice to avoid heap allocation with the vector no?

```suggestion
        let zi_packed = Ext::ExtensionPacking::from(zi);
```

src/poly/evals.rs
```diff
@@ -65,27 +221,139 @@ where
...
+
+        // Accumulate eq for remaining variables in packed form
+        let mut packed = Packed::zero_vec(1 << (n - n_pack));
+        eq_serial(
+            &mut packed,
+            &point.get_subpoint_over_range(..n - n_pack),
+            Packed::from_ext_slice(&init),
+        );
+
+        Self(packed)
```
tcoratger: Can't we speedup by using a parallel approach for the prefixes not sure if this worth it?


```suggestion
        // Threshold for switching from serial to parallel processing. Not sure about this threshold?
        const PARALLEL_THRESHOLD: usize = 1024;

        let n = point.num_variables();
        assert_ne!(scale, EF::ZERO);
        let n_pack = log2_strict_usize(F::Packing::WIDTH);
        assert!(n >= n_pack);

        // COMPUTE SUFFIX (Inside the SIMD lanes) 
        // 
        // We compute the equality polynomial for the last `n_pack` variables.
        // This forms a single `Packed` element which acts as the "seed" for the next stage.
        let seed = {
            // We use a temporary vector for the unpacked suffix calculation.
            let width = 1 << n_pack;
            let mut data = unsafe { uninitialized_vec::<EF>(width) };
            
            // Base case: Initialize first element
            data[0] = scale;

            // Expand sequentially for the suffix variables
            let suffix = point.get_subpoint_over_range(n - n_pack..);
            for (i, &var) in suffix.iter().rev().enumerate() {
                let count = 1 << i;
                let (lo, hi) = data.split_at_mut(count);
                // Serial loop is faster here due to small size
                lo.iter_mut().zip(hi.iter_mut()).for_each(|(l, h)| {
                    let product = *l * var;
                    *h = product;
                    *l -= product;
                });
            }
            Packed::from_ext_slice(&data)
        };

        // COMPUTE PREFIX (Vector Expansion) 
        // 
        // We expand the seed across the remaining variables using Packed arithmetic.
        let prefix_len = n - n_pack;
        let num_packed_elements = 1 << prefix_len;

        // Allocation: Use uninitialized memory to avoid memset overhead.
        let mut packed_data = unsafe { uninitialized_vec::<Packed>(num_packed_elements) };
        
        // Initialize the first element with the seed computed above
        packed_data[0] = seed;

        let prefix = point.get_subpoint_over_range(..n - n_pack);
        
        // Iterate backwards through the prefix variables to expand the hypercube
        for (i, &var) in prefix.iter().rev().enumerate() {
            let count = 1 << i;
            
            // We split the buffer into the "low" half (already initialized)
            // and the "high" half (uninitialized destination).
            //
            // Logic:
            // High[k] = Low[k] * var
            // Low[k]  = Low[k] * (1 - var)  => Low[k] - High[k]
            let (lo, hi_rest) = packed_data.split_at_mut(count);
            let hi = &mut hi_rest[..count];

            // For large chunks, we parallelize.
            if count >= PARALLEL_THRESHOLD {
                lo.par_iter_mut()
                    .zip(hi.par_iter_mut())
                    .for_each(|(l, h)| {
                        let product = *l * var;
                        *h = product;
                        *l -= product;
                    });
            } else {
                lo.iter_mut()
                    .zip(hi.iter_mut())
                    .for_each(|(l, h)| {
                        let product = *l * var;
                        *h = product;
                        *l -= product;
                    });
            }
        }

        Self(packed_data)
```

src/poly/multilinear.rs
```diff
@@ -14,6 +19,18 @@ use crate::poly::evals::EvaluationsList;
...
+    fn from(vars: Vec<F>) -> Self {
+        Self::new(vars)
+    }
+}
+
+impl<F: Field> From<&[F]> for MultilinearPoint<F> {
+    fn from(vars: &[F]) -> Self {
+        Self::new(vars.to_vec())
+    }
+}
```
tcoratger: Ideally, we don't want these methods for the following reason. That's perfectly acceptable, of course, but we'd like the user to realize they're creating an object that represents a point via their coordinates, and most of the time, `From` calls are an easy way to convert things without the user truly understanding what they're doing.

I'd prefer the user to have to call the `new` method so they realize they are playing with coordinates, but if this is absolutely necessary, then we can do that as well; it's not a blocker for merging.

src/sumcheck/sumcheck_single.rs
```diff
@@ -83,6 +256,55 @@ where
...
+    let evals = evals.compress_packed(r);
+
+    // Update sum: h(r) = c_0 + c_1 * r + c_2 * r^2 where c_1 = h(1) - c_0 - c_2
+    // Since h(1) = claimed_sum - h(0) = claimed_sum - c_0, we have c_1 = claimed_sum - 2*c_0 - c_2
+    // So h(r) = c_2 * r^2 + (claimed_sum - 2*c_0 - c_2) * r + c_0
+    let h_1 = *sum - c_0;
+    *sum = c_2 * r.square() + (h_1 - c_0 - c_2) * r + c_0;
+
+    (r, evals)
+}
```
tcoratger: Maybe in the Quad implementation we could have something like (in order to avoid having both `initial_round_packed` and `initial_round`)

```rust
/// Special handling for the initial round where evals are in Base Field F
    /// and weights are in Extension Field EF.
    #[instrument(skip_all)]
    pub(crate) fn initial_round_from_base<Challenger>(
        sumcheck_data: &mut SumcheckData<EF, F>,
        challenger: &mut Challenger,
        base_evals: &EvaluationsList<F>,
        weights: &mut EvaluationsList<EF>, // Assuming passed as EF, we might pack internally
        sum: &mut EF,
        pow_bits: usize,
    ) -> (EF, Self)
    where
        Challenger: FieldChallenger<F> + GrindingChallenger<Witness = F>,
    {
        // Check if we should operate in Packed mode
        let use_packed = base_evals.num_variables() > PACK_THRESHOLD;

        let (c_0, c_2) = if use_packed {
            let packed_evals = F::Packing::pack_slice(base_evals.as_slice());
            let packed_weights = EF::ExtensionPacking::pack_slice(weights.as_slice());
            
            let (t0, t2) = compute_sumcheck_coefficients(packed_evals, packed_weights);
            (
                EF::ExtensionPacking::to_ext_iter([t0]).sum::<EF>(),
                EF::ExtensionPacking::to_ext_iter([t2]).sum::<EF>(),
            )
        } else {
            compute_sumcheck_coefficients(base_evals.as_slice(), weights.as_slice())
        };

        // Transcript & Grinding
        sumcheck_data.polynomial_evaluations.push([c_0, c_2]);
        challenger.observe_algebra_element(c_0);
        challenger.observe_algebra_element(c_2);
        sumcheck_data.push_pow_witness(pow_grinding(challenger, pow_bits));

        let r: EF = challenger.sample_algebra_element();

        // Compress and Transition to Quad
        // Here we perform the Base -> Extension compression transition
        let quad = if use_packed {
            // Compress Base -> Packed Ext
            let new_evals = base_evals.compress_packed(r);
            let weights_packed = weights.compress_packed(r); 
            
            Self::Packed { evals: new_evals, weights: weights_packed }
        } else {
            let new_evals = base_evals.compress_ext(r);
            weights.compress(r);
            Self::Small { evals: new_evals, weights: weights.clone() }
        };

        // Update Sum
        let h_1 = *sum - c_0;
        *sum = c_2 * r.square() + (h_1 - c_0 - c_2) * r + c_0;

        (r, quad)
    }
```

src/whir/constraints/statement/eq.rs
```diff
@@ -1,12 +1,90 @@
...
+fn flat_eqs<F: Field, EF: ExtensionField<F>>(
+    points: RowMajorMatrixView<'_, EF>,
+    alpha: EF,
+) -> Vec<EF> {
+    let k = points.height();
+    let n = points.width();
+    assert_ne!(n, 0);
+
+    let mut acc = EF::zero_vec(n * (1 << k));
+    acc[..n].copy_from_slice(&alpha.powers().take(n).collect());
```
tcoratger:
```suggestion
    acc[..n].copy_from_slice(&alpha.powers().collect_n(n));
```

src/whir/constraints/statement/eq.rs
```diff
@@ -1,12 +1,90 @@
...
 use p3_multilinear_util::eq_batch::eval_eq_batch;
+use p3_util::log2_strict_usize;
 use tracing::instrument;
 
 use crate::poly::{evals::EvaluationsList, multilinear::MultilinearPoint};
 
+/// Builds the eq coefficients for every point in batch.
+/// Coefficients interleaved by list index: [a0, b0, c0, ..., a1, b1, c1, ...].
+/// `flat_eqs` also weights point i by alpha^i
+fn flat_eqs<F: Field, EF: ExtensionField<F>>(
```
tcoratger: I imagine this is not costly at all that's why we don't use any parallelization strategy?

src/poly/evals.rs
```diff
@@ -845,7 +1055,7 @@ impl<F> IntoIterator for EvaluationsList<F> {
 /// - `evals`: A slice containing the `2^n` evaluations of the polynomial over the boolean
 ///   hypercube, ordered lexicographically. For `n=2`, the order is `f(0,0), f(0,1), f(1,0), f(1,1)`.
 /// - `point`: A slice containing the `n` coordinates of the point `p` at which to evaluate.
-fn eval_multilinear<F, EF>(evals: &[F], point: &MultilinearPoint<EF>) -> EF
+pub fn eval_multilinear<F, EF>(evals: &[F], point: &MultilinearPoint<EF>) -> EF
```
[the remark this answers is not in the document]
tcoratger: Yes makes sense but as we do not use this function at all in the code (if I don't miss anythin), surely we must delete it to avoid having an unused function that is just called in the benchmarks.

My recommendation is the following: once your benchmarks show that the new functions introduced are ok in terms of cost when compared to this one, we can completely remove this function and so all the related calls in the benchmarks, the new functions being our new reference points.

src/utils.rs
```diff
@@ -53,33 +16,92 @@ pub unsafe fn uninitialized_vec<A>(len: usize) -> Vec<A> {
...
+
+/// Unpack packed extension field elements to the standard representation.
+#[inline]
+pub fn unpack_slice<F: Field, Ext: ExtensionField<F>>(
+    packed: &[Ext::ExtensionPacking],
+) -> Vec<Ext> {
+    let mut out = Ext::zero_vec(packed.len() * F::Packing::WIDTH);
+    unpack_slice_into(&mut out, packed);
+    out
+}
```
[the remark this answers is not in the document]
kilic: Method in p3 only applies for base field so I had to introduce this one

src/utils.rs
```diff
@@ -53,33 +16,92 @@ pub unsafe fn uninitialized_vec<A>(len: usize) -> Vec<A> {
...
+
+#[inline]
+/// Pack extension field elements into their packed representation.
+pub fn pack_slice<F: Field, Ext: ExtensionField<F>>(slice: &[Ext]) -> Vec<Ext::ExtensionPacking> {
```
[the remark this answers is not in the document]
kilic: Same issue with above

src/sumcheck/sumcheck_single.rs
kilic: I've tackled with this issue somehow similar way suggested above

src/utils.rs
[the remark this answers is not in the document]
tcoratger: Could it help?

https://github.com/Plonky3/Plonky3/pull/1180

src/poly/multilinear.rs
```diff
@@ -341,6 +364,35 @@ where
...
+        });
+        RowMajorMatrix::new(flat, n)
+    }
+
+    /// Given a position splits the point into two sub-points.
+    pub(crate) fn split_at(&self, pos: usize) -> (Self, Self) {
+        (
+            Self::new(self.0.split_at(pos).0.to_vec()),
+            Self::new(self.0.split_at(pos).1.to_vec()),
+        )
```
tcoratger:
```suggestion
        let (left, right) = self.0.split_at(pos);
        (Self::new(left.to_vec()), Self::new(right.to_vec()))
```

src/poly/evals.rs
```diff
@@ -910,106 +1077,177 @@ where
...
-                            .zip_eq(left.iter())
-                            .map(|(&a, &b)| b * a)
-                            .sum::<EF>()
-                            * c
-                    })
-                    .sum()
-            } else {
-                // Create a new point with the remaining coordinates.
-                let sub_point = MultilinearPoint::new(tail.to_vec());
+pub fn eval_multilinear_recursive<F, EF>(evals: &[F], point: &MultilinearPoint<EF>) -> EF
```
tcoratger: For this method I think that if we manage to minimize internal allocations, given that the recursive calls can really hurt performance, we could make huge performance gains.

src/sumcheck/sumcheck_single.rs
```diff
@@ -1,150 +1,256 @@
...
+        match self {
+            Self::Packed { evals, .. } => evals.num_evals() * F::Packing::WIDTH,
+            Self::Small { evals, .. } => evals.num_evals(),
+        }
+    }
+
+    fn transition(&mut self) {
+        let k = self.k();
+        match self {
+            Self::Packed { evals, weights } if k < PACK_THRESHOLD => {
```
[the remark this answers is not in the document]
kilic: In one of the last commits I changed it so packing threshold is calculated dynamically

src/sumcheck/tests.rs
```diff
@@ -1048,14 +1048,15 @@ fn test_sumcheck_prover_svo() {
 fn test_sumcheck_prover_with_skip() {
     let mut rng = SmallRng::seed_from_u64(0);
 
-    for num_vars in 8..=10 {
+    for num_vars in 10..=10 {
         for folding_factor in 2..=num_vars {
-            for _ in 0..100 {
+            for _ in 0..1 {
```
SyxtonPrime: Not a problem but is there a reason you've cut down on the test size?

src/poly/evals.rs
```diff
@@ -910,96 +1065,158 @@ where
...
-                        // For smaller subproblems, execute sequentially.
-                        (
-                            eval_multilinear(f0, &sub_point),
-                            eval_multilinear(f1, &sub_point),
-                        )
-                    }
-                };
-                // Perform the final linear interpolation for the first variable `x`.
-                f0_eval + (f1_eval - f0_eval) * *x
-            }
```
SyxtonPrime: It's a little unclear to me why you've pulled `eval_multilinear_recursive` out of `eval_multilinear_small`.

It seems like you could put all of the code of `eval_multilinear_recursive` into the `_` part of the match statement of `eval_multilinear_small`.

src/sumcheck/sumcheck_single.rs
```diff
@@ -152,95 +279,29 @@ where
...
 #[instrument(skip_all, level = "debug")]
-fn compute_sumcheck_coefficients<F: Field, EF: ExtensionField<F>>(
-    evals: &EvaluationsList<F>,
-    weights: &EvaluationsList<EF>,
-) -> (EF, EF) {
-    assert!(evals.num_variables() >= 1);
+fn compute_sumcheck_coefficients<A, B>(evals: &[A], weights: &[B]) -> (B, B)
+where
+    A: Copy + Send + Sync + Algebra<A>,
+    B: Copy + Send + Sync + Algebra<A> + Algebra<B>,
```
SyxtonPrime: Requiring something be an algebra over itself is a bit meaningless. I think the "right" trait bounds here are:
```suggestion
    A: Copy + Send + Sync + PrimeCharacteristicRing,
    B: Copy + Send + Sync + Algebra<A>,
```
Note that being an algebra implies you are a `PrimeCharacteristicRing`.

PR #385, opened 2025-12-04
serealization skip

camofu:
skip serializing default values: `using F::default()` instead of `Option`, we're putting zeros into the
  proof even when `pow_bits == 0`. With this PR, we avoid serializing it into the proof.


tcoratger (approved):
Looks good, I think in practice this almost never happens because we almost always have a small amount of grinding but I agree that it never hurt to add this.

Can we just have a small unit test to confirm that:
- With POW the witness is properly serialized
- Without POW this is not serialized

Issue #338, opened 2025-11-07
sumcheck svo: full integration


tcoratger:
Currently, we have the implementation for paper (https://eprint.iacr.org/2025/1117.pdf), which has been merged here (https://github.com/tcoratger/whir-p3/pull/322).

However, this is not currently applied to the WHIR protocol. To address this, an enum to handle various sumcheck optimizations for the first round has been introduced here (https://github.com/tcoratger/whir-p3/pull/335) to facilitate user selection.

So it would be good to completely integrate the SVO to the protocol by replacing this fallback to the classic sumcheck

https://github.com/tcoratger/whir-p3/blob/5981a1b62e23edc7b721ed55a1f90f7c3919cc52/src/whir/prover/round_state/state.rs#L184-L195

by a call to the optimized SVO version (should be `SumcheckSingle::from_base_evals_svo` instead of `SumcheckSingle::from_base_evals` I guess).

For this to happen, I think that we should fix this (and maybe other things, to be confirmed)

https://github.com/tcoratger/whir-p3/blob/5981a1b62e23edc7b721ed55a1f90f7c3919cc52/src/sumcheck/sumcheck_single_svo.rs#L43-L46

as well.

In https://github.com/tcoratger/whir-p3/pull/337, an end to end test has been added with `SumcheckOptimization::Svo` but it just falls back on the classic sumcheck currently due to the previously mentioned element. After solving this issue, this end to end test should properly test the SVO optimization.

tcoratger:
See this document for more information: https://hackmd.io/@tcoratger/H1SNENAeZg

PR #388, opened 2025-12-07
integrate SVO optimization in WHIR prover

adust09:
related to #338

### DIFF
--- a/src/whir/prover/round_state/state.rs
+++ b/src/whir/prover/round_state/state.rs
@@ -172,11 +172,17 @@ where
                 )
             }
 
-            // Branch: WithStatementSvo - SVO optimization (Algorithm 6 from https://eprint.iacr.org/2025/1117)
+            // Branch: WithStatementSvo - SVO optimization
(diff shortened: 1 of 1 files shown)


src/whir/prover/round_state/state.rs
```diff
@@ -172,22 +172,44 @@ where
                 )
             }
 
-            // Branch: WithStatementSvo - SVO optimization (currently falls back to classic)
+            // Branch: WithStatementSvo - SVO optimization (Algorithm 6 from https://eprint.iacr.org/2025/1117)
             InitialPhase::WithStatementSvo { sumcheck } => {
+                // SVO optimization requirements:
+                // 1. At least 2 * NUM_SVO_ROUNDS (6) variables (NUM_SVO_ROUNDS = 3)
```
tcoratger: Just to avoid noting any numbers in case we want to change in the future. Also could be nice to explain why we need this constraint from the paper?

```suggestion
                // 1. At least 2 * NUM_SVO_ROUNDS variables
```

PR #394, opened 2025-12-12
primitives: rm coeff and wavelet

### LINKED ISSUE #391: primitives: remove coeff and wavelet files
These files 

https://github.com/tcoratger/whir-p3/blob/main/src/poly/coeffs.rs

and

https://github.com/tcoratger/whir-p3/blob/main/src/poly/wavelet.rs

are not used anywhere in the code except for testing. 

We should remove them and rely on the evaluation format for testing purposes everywhere the coefficient form was used.

ThreeHrSleep:
should close #391


tcoratger:
I think that you can remove the wavelet benchmark file as well :)

Issue #228, opened 2025-07-26
[LONG TERM] Batch opening


TomWambsgans:
In the long term, it would be nice to support batch opening.

Some PIOP (typically lookups) require to commit to a polynomial, receive a random challenge, and then to commit to a second polynomial (potentially bigger / smaller). At the end of the protocol, we need to open the 2 polynomials at 2 different points. It should be doable to modify the protocol to allow an efficient "batch opening" (where the proof would be smaller than the concatenation of the 2 individual opening proofs).

tcoratger:
@CarloModicaPortfolio Can you share the document here for @kilic to have a look?

CarloModicaPortfolio:
@kilic here:
https://hackmd.io/b4Tx6F00QUGMWiPl33vdYQ

PR #397, opened 2025-12-23
batch proving with same number of variables - part 1

camofu:
Addressing issue: 

https://github.com/tcoratger/whir-p3/issues/228

### DIFF
--- a/src/whir/batch_proof.rs
+++ b/src/whir/batch_proof.rs
@@ -29,6 +29,7 @@
 //! The protocol then continues with standard WHIR on the folded polynomial.
 
 use alloc::{vec, vec::Vec};
+use hashbrown::Equivalent;
 use p3_challenger::{FieldChallenger, GrindingChallenger};
 use p3_dft::TwoAdicSubgroupDft;
 use p3_field::{ExtensionField, Field, TwoAdicField};
@@ -344,7 +345,8 @@ where
         proof: &BatchWhirProof<F, EF, DIGEST_ELEMS>,
         challenger: &mut Challenger,
         parsed_commitments: Vec<&ParsedCommitment<EF, Hash<F, F, DIGEST_ELEMS>>>,
-        mut statement: EqStatement<EF>,
+        mut statement_a: EqStatement<EF>,
+        mut statement_b: EqStatement<EF>,
     ) -> Result<MultilinearPoint<EF>, VerifierError>
     where
         H: CryptographicHasher<F, [F; DIGEST_ELEMS]> + Sync,
--- a/src/whir/prover/mod.rs
+++ b/src/whir/prover/mod.rs
@@ -138,8 +138,7 @@ where
     ///
     ///
     /// # Errors
-    /// Returns an error if the witness or statement are invalid, or if a round fails.
-    #[instrument(skip_all)]
+    /// Returns an error if the witness or statement are invalid, or if a round f   #[instrument(skip_all)]
     pub fn prove<Dft: TwoAdicSubgroupDft<F>, const DIGEST_ELEMS: usize>(
         &self,
         dft: &Dft,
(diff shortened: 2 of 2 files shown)


src/poly/evals.rs
```diff
@@ -919,6 +919,23 @@ where
...
+
+    /// Compute g = coeff_a·f_a + coeff_b·f_b element-wise
+    pub fn linear_combination<EF: ExtensionField<F>>(
+        f_a: &Self,
+        coeff_a: EF,
+        f_b: &Self,
+        coeff_b: EF,
+    ) -> EvaluationsList<EF> {
+        assert_eq!(f_a.num_evals(), f_b.num_evals());
+        let evals: Vec<EF> = f_a
```
tcoratger:
```suggestion
        let evals = f_a
```

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+//! ```text
+//! f_c(X, x_1, ..., x_m) = X·f_a(x_1, ..., x_m) + (1-X)·f_b(x_1, ..., x_m)
+//! ```
+//!
+//! The selector variable X chooses between f_a (when X=1) and f_b (when X=0).
+//!
+//! The combined weight polynomial is:
+//!
+//! ```text
+//! w(X, b) = X·eq(b, z_a) + α(1-X)·eq(b, z_b)
```
tcoratger: Here I think that you need to explain what alpha is because this was never mentioned before in the doc.

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Commitment to second polynomial (f_b)
+    pub commitment_b: [F; DIGEST_ELEMS],
+
+    /// Selector sumcheck data: stores [c0, c2] for h(X) = c0 + c1·X + c2·X²
+    /// c0 = h(0) = α·v_b
+    /// c2 = quadratic coefficient
+    /// Verifier derives h(1) = σ - h(0) = v_a
+    pub selector_sumcheck: SumcheckData<EF, F>,
+
+    /// Inner WHIR proof on the folded polynomial g = r_0·f_a + (1-r_0)·f_b
```
tcoratger: Missing a alpha?

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        // ==== Step 1: Store commitments ====
+        proof.commitment_a = witness_a.prover_data.root().into();
+        proof.commitment_b = witness_b.prover_data.root().into();
+
+        // ==== Step 2: Sample batching randomness α ====
+        let alpha: EF = challenger.sample_algebra_element();
+
+        // ==== Step 3: Extract claims from statements ====
+        // For now, we assume single-constraint statements
```
tcoratger: What happens if we have multiple constraints in the statements? which is usually the case? A TODO for the future or do you plan to solve it here?

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+        [F; DIGEST_ELEMS]: Serialize + for<'de> Deserialize<'de>,
+    {
+        // Validate that both polynomials have the same number of variables
+        assert_eq!(
+            witness_a.polynomial.num_variables(),
+            witness_b.polynomial.num_variables(),
+            "Batch opening requires same-degree polynomials"
+        );
+
+        // ==== Step 1: Store commitments ====
```
tcoratger: Very small nit but could we remove all the "====" stuffs from the comments, just for consistency with the rest of the codebase?

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        (sumcheck_prover, r_0)
+    }
+}
+
+/// Extracts a single constraint (point, evaluation) from an EqStatement.
+///
+/// # Panics
+///
+/// Panics if the statement is empty.
```
tcoratger: Probably we need a TODO here to say that we have to remove it no? because in the general case we should not need this since the protocol is not fixed to a single constraint right?

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+
+        // Verify σ = h(0) + h(1) = v_a + α·v_b
+        let sigma = v_a + alpha * v_b;
+        let h_1 = sigma - c0;
+        assert_eq!(h_1, v_a);
+    }
+
+    /// Test that folding produces the correct polynomial.
+    #[test]
+    fn test_folding() {
```
tcoratger: A test like this seems to test only the `linear_combination` function right? So it should be moved to the evals.rs file no?

tcoratger (commented):
Just left a couple of comments. I'm not sure about the usefulness of the added tests. 

They don't seem to directly test the functions in the file, but rather similar behaviors, which is pretty weird because then we don't have any complete tests calling `batch_prove` or `selector_round` functions.

src/whir/batch_proof.rs
CarloModicaPortfolio: Resolved with commut 91e3a70

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,456 @@
...
+            assert_eq!(g.as_slice()[i], expected);
+        }
+    }
+
+    /// Test that the folded sum matches h(r_0) from the sumcheck polynomial.
+    ///
+    /// This verifies the key sumcheck invariant: after folding with challenge r_0,
+    /// the new sum σ' = Σ_b g(b)·w'(b) equals h(r_0).
+    #[test]
+    fn test_folded_sum() {
```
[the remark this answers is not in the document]
CarloModicaPortfolio: Again, sorry removed here 3d7be34

CarloModicaPortfolio:
You are right the tests were ill-formed. I added a new test for `selector_round`, however I have not added one for batch prove as the function is still in progress. In the meantime I also added the folded OOD constraints, using the same points for both polynomials

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,423 @@
...
+//!
+//! Given two polynomials f_a and f_b (both with m variables), we construct a combined
+//! polynomial f_c with m+1 variables:
+//!
+//! ```text
+//! f_c(X, x_1, ..., x_m) = X·f_a(x_1, ..., x_m) + α(1-X)·f_b(x_1, ..., x_m)
+//! ```
+//!
+//! The selector variable X chooses between f_a (when X=1) and f_b (when X=0).
+//! α is the folding randomness chocen by the Verifier.
```
tcoratger:
```suggestion
//! α is the folding randomness chosen by the Verifier.
```

src/whir/batch_proof.rs
tcoratger: No of course I have no problem to merge it as fast as possible. Even on the contrary, I prefer that we merge a first iteration with polynomials of same size and single constraint so that we can have a first idea and then iterate on the rest with followup PRs.

But I would still like us to have an end to end test with single constraint and for example 2 polynomials of the same size where the prove/verify workflow is working so that we are sure what we are doing is fully right.

Does this make sense?

My goal here is to make sure we don't merge code that is false and not working. In our case I think that the only way to confirm everything is working fine is by running the full end to end test.

CarloModicaPortfolio: Yes it does makes sense to me. I agree that to make sure the base case is correct is to run the prove/verify workflow, it does not make much sense to test (and therefore merge) uncomplete functions. 🎄

src/whir/batch_proof.rs
```diff
@@ -0,0 +1,423 @@
...
+        let pow_bits = self.starting_folding_pow_bits;
+        let r_0 = selector_data.observe_and_sample::<Challenger, F>(challenger, c0, c2, pow_bits);
+
+        // Fold the polynomial and weights:
+        // g(b) = r_0·f_a(b) + (1-r_0)·f_b(b)
+        // w'(b) = r_0·eq(b, z_a) + α(1-r_0)·eq(b, z_b)
+        // σ' = h(r_0)
+
+        // Folded polynomial in extension field
+        let g = EvaluationsList::linear_combination(f_a, r_0, f_b, EF::ONE - r_0);
```
kilic: Similarly maybe it is possible to combine f_a and f_b onto f_a by getting f_a with mut value

Issue #305, opened 2025-10-03
Fiat Shamir: rm current implementation to switch to Plonky3


tcoratger:
### Description

The current implementation of the Fiat-Shamir transform can be found here: https://github.com/tcoratger/whir-p3/tree/main/src/fiat_shamir

But it deviates from the standard practice established in Plonky3. It currently uses a "streaming" approach where data is transmitted from the prover to the verifier via a large `proof_data` vector within the Fiat-Shamir mechanism itself.

This is problematic because Plonky3's design philosophy mandates that the Fiat-Shamir transcript should be used exclusively for generating randomness by absorbing proof data, not for the transmission of the data itself.

### Proposed Solution

To align with Plonky3's architecture, we need to refactor our proof structure. All data currently sent through the custom Fiat-Shamir `proof_data` vector should instead be transmitted via a dedicated proof struct.

### Action Items

1.  **Deprecate the Custom Fiat-Shamir Implementation**: The entire directory at https://github.com/tcoratger/whir-p3/tree/main/src/fiat_shamir should be removed. We will rely solely on Plonky3's Fiat-Shamir for randomness.

2.  **Create a `WhirProof` Struct**: We must introduce a new, explicit proof structure (e.g., `WhirProof`) to bundle all the data that the prover sends to the verifier. This follows the pattern seen in other Plonky3 components, such as FRI's `FriProof` struct: https://github.com/Plonky3/Plonky3/blob/2e6ff23090b5bd5138695ceb757181443a523c6d/fri/src/proof.rs#L7-L17

3.  **Refactor Data Handling**:

      * Identify all data currently being pushed into the `proof_data` vector.
      * Store this data within fields in the new `WhirProof` struct instead.
      * While the data will be transmitted via the `WhirProof` struct, it must still be *absorbed* by the Plonky3 Fiat-Shamir transcript to ensure the verifier's challenges are cryptographically bound to the proof data.

#### Example: Sumcheck Evaluations

As a concrete example, the sumcheck polynomial evaluations found here:
https://github.com/tcoratger/whir-p3/blob/ddea7f482f8299fda2705fbdbf8e9d5ee065b5bf/src/sumcheck/sumcheck_single.rs#L50-L52

Under the new design:

  - The `evaluations` would be stored in a field within the `WhirProof` struct.
  - The prover would still pass `evaluations` to the Plonky3 Fiat-Shamir transcript to contribute to its state for generating subsequent challenges.
[...]

PR #402, opened 2026-01-01
Serializing transcript

kilic:
This PR adds a global and serializing transcript and removes `WhirProof`

Currently in a fiat shamir stream, a mutable proof object and a p3-challenger has to move and operate together. New transcript object wraps them into single object `FiatShamirWriter` for prover side and `FiatShamirReader` for verifier side. New transcript immediately serializes (or deserialize) elements in proof so it should work with any kind of proof stream. For future improvements we can add IO pattern enforcement.

### DIFF
--- a/src/fiat_shamir/transcript.rs
+++ b/src/fiat_shamir/transcript.rs
@@ -376,6 +376,7 @@ where
 
 #[cfg(test)]
 mod test {
+
     use p3_baby_bear::{BabyBear, Poseidon2BabyBear};
     use p3_challenger::{DuplexChallenger, FieldChallenger, GrindingChallenger};
     use p3_field::{ExtensionField, Field, extension::BinomialExtensionField};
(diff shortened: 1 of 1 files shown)


tcoratger:
Indeed I think that this approach simplifies the code quite a lot because this removes various structures. The information now passes via the Fiat Shamir object, reducing the number of needed alternative objects to represent stuffs.

However I would still like a second opinion from @SyxtonPrime about this. We started the effort with this issue https://github.com/tcoratger/whir-p3/issues/305 after the workshop in Cambridge in order to facilitate Plonky3 integration in the relatively near future.

Via the proof objects we were able to approach very closely to what is done traditionally in other Plonky3 similar algorithms like FRI.

Of course I consider the transcript thing as something more secure and easier to not make mistakes but this deviates a bit from the Plonky3 way of doing things. And I would like that not to prevent us from upstream WHIR easily in Plonky3.

What do you think @SyxtonPrime? Maybe having something like the `transcript.rs` file proposed in this PR could be useful in Plonky3, don't you think? So that everything in Plonky3 would go through a transcript?

PR #403, opened 2026-01-05
sumcheck svo: full integration

tcoratger:


src/sumcheck/sumcheck_small_value.rs
```diff
@@ -430,6 +439,128 @@ pub fn algorithm_5<Challenger, F, EF>(
...
+    //
+    // Now p' is in extension field, and we continue with standard sumcheck.
+    let mut folded_evals: EvaluationsList<EF> = poly.compress_ext(r_1);
+    let mut folded_weights: EvaluationsList<EF> = weights.clone();
+    folded_weights.compress(r_1);
+
+    // Rounds 2-3: Standard Sumcheck
+    //
+    // Use standard sumcheck on the folded polynomials.
+    // This correctly handles the cross-terms.
```
tcoratger: @quangvdao Here I've tried a first attempt at fully integrating SVO optimization with multiple constraints to WHIR, as discussed in detail in Option 2 of https://hackmd.io/@tcoratger/H1SNENAeZg.

Due to cross-terms after Round 1 with the folding and accumulators, I'm not sure how to apply the pre-computed accumulators for Rounds 2 and 3. Is it even possible to do this, or should we only apply SVO to Round 1 and then switch to normal sumcheck for the rest of the rounds (in which case I'm not sure about the optimization's effectiveness)?

Let me try to summarize a bit my understanding here.

We want to optimize the sumcheck prover for a **batched** statement involving a polynomial $p(x)$ (small values) and a weight polynomial $W(x)$ (large extension field values), where $W$ is a linear combination of equality constraints:
$$W(x) = \sum \gamma_i \cdot eq(z_i, x)$$

**My initial optimization attempt with SVO**
The naive optimization attempts to precompute accumulators for all rounds $k$ upfront (3 rounds for example). This assumes that the folded result of round $k$ can be derived solely from the accumulators of round $k-1$.

**The Math**
In Round 1, we fold variable $x_0$ with challenge $r$. The new polynomials $W'$ and $p'$ become:
$$W'(x') = (1-r)W(0, x') + r W(1, x')$$
$$p'(x') = (1-r)p(0, x') + r p(1, x')$$

For Round 2, we need to compute the sum over the product of these folded polynomials:
$$\sum_{x'} W'(x') \cdot p'(x')$$

Expanding this product reveals the issue:
$$W' \cdot p' = \big((1-r)W_0 + r W_1\big) \cdot \big((1-r)p_0 + r p_1\big)$$
$$= (1-r)^2 (W_0 \cdot p_0) + r^2 (W_1 \cdot p_1) + r(1-r)(W_0 \cdot p_1 + W_1 \cdot p_0)$$

where $r(1-r)(W_0 \cdot p_1 + W_1 \cdot p_0)$ are cross terms.


**The Issue**
* **Single Constraint:** If $W = eq(z, x)$, it factorizes cleanly ($W_0 \propto W_1$), causing the cross terms to collapse into the main terms.
* **Batched Constraints:** $W$ does **not** factorize. $W_0$ and $W_1$ are distinct linear combinations.
* **Result:** Precomputed accumulators only track $W_0 p_0$ and $W_1 p_1$. They fundamentally lack the information to reconstruct the cross terms ($W_0 p_1 + W_1 p_0$).


That's why is this draft PR I've tried first to apply the pre computed accumulators only to the first round and then switch to basic sumcheck. But then the optimization looks pretty unefficient no?

tcoratger:
Fixed by #409

